PR #7, opened 2021-08-11
🐛 [i18n] Fix broken routes

nwingt:


williamchong:
Should we use `no_prefix` and manually control i18n in a page instead?

nwingt:
Seems they don't support controlling locale with query string (although we can implement by ourselves) but I think using sugar functions from nuxt-i18n are fine

PR #42, opened 2021-08-31
🐛 Fix disable behavior

AuroraHuang22:
Fix disable behavior issue from [here](https://www.notion.so/likecoin/db890d07b1484dc3bd7f45bcfdc8f08e)


nwingt (commented):
應該在 Button 處理比較好？

nwingt (approved):

PR #43, opened 2021-09-01
✨ [/new/index] Able to upload multiple types of files

AuroraHuang22:
Fix issues raised from [here](https://www.notion.so/likecoin/should-allow-text-files-to-be-uploaded-da152527cc0540558acdc0f5ed23d70f)

1. Add feature to upload multiple types of files in `<UploadForm/>`
2. Modify `<UploadForm/>` `<IscnRegisterForm/>` `<IscnUploadedInfo/>` layout
3. Add File icon


components/UploadForm.vue
```diff
@@ -67,15 +67,16 @@
             <input
               ref="imageFile"
               class="hidden"
-              accept="image/*"
+              accept="image/*,audio/*,video/*,.doc,.docx,.pdf,txt,"
```
AuroraHuang22, line 70: 還是不寫 accept？ 這樣就能支援所有檔案類型

williamchong: 可以不寫 accept

nwingt (changes requested):
I see that <IconFile /> is repeating, can be refactorized

components/UploadForm.vue
```diff
@@ -230,7 +218,11 @@ export default class UploadForm extends Vue {
         // eslint-disable-next-line prefer-destructuring
         this.fileBlob = files[0]
         if (this.isImage) {
-          this.exifInfo = await exifr.parse(files[0])
+          try {
+            this.exifInfo = await exifr.parse(files[0])
+          } catch (exifInfo) {
+            this.exifInfo = null
```
williamchong, line 224:
```suggestion
          } catch (err) {
            console.error(err)
            this.exifInfo = null
```
I think it is better to log the error just in case

williamchong (changes requested):
Do we have any UI in the upload flow/view page to display the file type of a non-image file?

williamchong (approved):

PR #46, opened 2021-09-07
✨ [Component] Add ISCN card

nwingt:
image

Todo:
- [x] Support responsive canvas
- [x] Include QR code
- [x] Change orientation


AuroraHuang22:
🤩🤩🤩

williamchong (approved):
I still think a looping animation would be cool

AuroraHuang22 (approved):

williamchong (approved):
Would be nice if we can have lazy load later in search results

pages/embed/card/_iscnId.vue
```diff
@@ -0,0 +1,104 @@
...
+  ) => Promise<{ records: ISCNRecordWithID[] }>
+
+  get record() {
+    return this.getISCNById(this.iscnId)
+  }
+
+  get isShowAdjustableProps() {
+    const { dev = '0' } = this.$route.query
+    return dev === '1'
+  }
```
williamchong, line 46: Should this be behind `NODE_ENV !== 'production'` ?

nwingt: This acts as an inspector, should come in handy even in production

PR #49, opened 2021-09-21
🐛 [TextField] Fix textarea unable to set value

nwingt:


components/TextField.vue
```diff
@@ -2,13 +2,10 @@
   <div class="flex flex-col">
     <component
       :is="tag"
-      v-bind="$attrs"
-      :class="rootClasses"
-      :value="value"
```
AuroraHuang22, line 7: I found that `:value.prop="value"` also works

nwingt: You mean for textarea?

AuroraHuang22: yes, I tried to only changed `:value="value"` to `:value.prop="value"` to solve the issue, and it works for me.

nwingt: Still not working
https://user-images.githubusercontent.com/17264716/134216207-9948ebb7-9129-4070-a2be-8d435807e09b.mov

nwingt: You may try setting a initial value for TextField rather than an empty string

AuroraHuang22 (approved):

PR #54, opened 2021-09-24
✨[UploadForm] Add warning to prevent users from connecting to Ledger

AuroraHuang22:


nwingt (approved):

components/UploadForm.vue
```diff
@@ -121,6 +121,25 @@
         </Dialog>
       </Card>
     </div>
+    <div
+      class="
+        min-w-[648px]
+        flex
+        items-center
+        justify-between
+        px-[43px]
```
nwingt, line 130: Should be design mistake ☺️
```suggestion
        px-[44px]
```

nwingt (approved):

PR #56, opened 2021-09-28
Update signing dialog UI

nwingt:


AuroraHuang22 (approved):

pages/new/index.vue
```diff
@@ -79,6 +83,10 @@ export default class NewIndexPage extends Vue {
     this.state = 'iscn';
   }
 
+  handleQuitRegistration() {
+    this.state = 'init';
```
williamchong, line 87: Should we also clear all local var?

PR #60, opened 2021-10-05
✨ Add feature to submit EXIF info to contentMetadata

AuroraHuang22:
- Upgrade iscn-js to version 0.0.6
- Save EXIF info into contentMetadata
- Improve related layout
- Reorder component options refer to [Vue](https://vuejs.org/v2/style-guide/index.html#Component-instance-options-order-recommended)


components/MetadataCard.vue
```diff
@@ -22,7 +23,7 @@
     </template>
     <FormField
       v-for="(item, name) in exifInfo"
-      :key="item.key"
+      :key="name"
       class="mb-[8px]"
       :label="name"
```
williamchong, line 28: Should consider using a i18n key here?

pages/view/_iscnId.vue
```diff
@@ -421,27 +384,103 @@ export default class ViewIscnIdPage extends Vue {
...
+            ? `${exifInfo.ExposureProgram}, 1/${(
+                1 / exifInfo.ExposureTime
+              ).toFixed(0)} s, f/${exifInfo.FNumber.toFixed(1)}, ISO ${
+                exifInfo.ISO
+              }`
+            : undefined,
+        Flash: exifInfo.Flash,
+        Date: exifInfo.CreateDate,
+        Size: exifInfo.Size,
+      }
```
williamchong, line 463: I think we should not use capitalized letter as object key, and should not assume this key will be directly display without i18n

williamchong (changes requested):

components/MetadataCard.vue
AuroraHuang22: how & why? 🤔

pages/view/_iscnId.vue
AuroraHuang22: So I just need to switch the key to i18n (but still as the capitalized letter)? Or maybe we can split each item as the getter instead of using an object to render?

components/MetadataCard.vue
```diff
@@ -57,8 +58,10 @@ export default class MetadataCard extends Vue {
   }
 
   mounted() {
-    if (this.data) this.exifInfo = this.data
-    else this.extractEXIFInfo()
+    if (this.data) {
+      this.exifInfo = this.data
+      this.isFromData = true
```
williamchong, line 63: Can `isFromData` be computed by `return !!this.data` instead?

PR #67, opened 2021-10-08
♻️ Update ISCN Uploading Page

AuroraHuang22:


components/IscnRegisterForm.vue
```diff
@@ -582,17 +582,22 @@ export default class IscnRegisterForm extends Vue {
   }
 
   get exif() {
-    const extension = this.fileType!.split('/')
-    return {
-      ...this.exifInfo,
```
williamchong, line 587: Do we still want to have information from `this.fileType` and  `this.fileSize` even if no exif?
If yes just guard the `...this.exifInfo` with empty object

components/IscnRegisterForm.vue
```diff
@@ -582,17 +582,22 @@ export default class IscnRegisterForm extends Vue {
...
+          this.type === 'Photo' || this.type === 'Image'
+            ? `${this.exifInfo.ExifImageHeight} x ${
+                this.exifInfo.ExifImageWidth
+              } ${extension[extension.length - 1].toUpperCase()} (${
+                this.fileSize
+              })`
+            : this.fileSize,
+      }
     }
+    return undefined
```
williamchong, line 600: Not sure if returning `{}` or `undefined` is more prefered

PR #68, opened 2021-10-11
✨ Support uploading of directory/file bundles

williamchong:
This PR allows uploading of more than 1 file, path identified by form field
This is useful for submitting website with relative path assets, or SPA built by nuxt/next/etc
Arweave part is implemented via [manifest](https://github.com/ArweaveTeam/arweave/blob/eee99d6ad3eb3da83399d1315ea3a68d9deaff88/doc/path-manifest-schema.md)
IPFS part is implemented via IPFS directory. Note that arweave-pifs has no way to pin the root IPFS hash, since it is a dag node instead of a file.


server/arweave/index.ts
```diff
@@ -28,38 +34,137 @@ export async function getArIdFromHashes(ipfsHash: string) {
...
+
+function generateManifestFile(files: ArweaveFile[], { stub = false } = {}): ArweaveFile {
+  const manifest = generateManifest(files, { stub });
+  return {
+    key: 'manifest',
+    mimetype: 'application/x.arweave-manifest+json',
+    buffer: Buffer.from(stringify(manifest), 'utf-8'),
+  };
+}
+async function uploadManifestFile(filesWithId: ArweaveFile[]) {
```
nwingt, line 76:
```suggestion

async function uploadManifestFile(filesWithId: ArweaveFile[]) {
```

server/arweave/index.ts
```diff
@@ -28,38 +34,137 @@ export async function getArIdFromHashes(ipfsHash: string) {
...
+  const manifest = generateManifestFile(filesWithId);
+  const manifestIPFSHash = await getFileIPFSHash(manifest);
+  let arweaveId = await getArIdFromHashes(manifestIPFSHash);
```
nwingt, line 79:
```suggestion
  let arweaveId = await getArweaveIdFromHashes(manifestIPFSHash);
```

server/arweave/index.ts
```diff
@@ -73,20 +178,87 @@ export async function submitToArweave(data: { mimetype: string; buffer: Buffer;
...
+  }
+
+  const [
+    folderIpfsHash,
+    ipfsHashes,
+  ] = await Promise.all([
+    getFolderIPFSHash(files),
+    Promise.all(files.map(f => getFileIPFSHash(f))),
+  ]);
+  const arweaveIds = await Promise.all(ipfsHashes.map(h => getArIdFromHashes(h)));
```
nwingt, line 212:
```suggestion
  const arweaveIds = await Promise.all(ipfsHashes.map(h => getArweaveIdFromHashes(h)));
```

server/arweave/router.ts
```diff
@@ -1,40 +1,90 @@
 import BigNumber from 'bignumber.js';
-import { Router } from 'express';
+import { NextFunction, Request, Response, Router } from 'express';
 import multer from 'multer';
-import { estimateARPrice, convertARPriceToLIKE, uploadFileToArweave } from '.';
+import { estimateARPrices, convertARPricesToLIKE, uploadFilesToArweave } from '.';
 import { getIPFSHash } from '../ipfs';
 import { queryLIKETransactionInfo } from '../like';
+import { ArweaveFile } from './types';
```
nwingt, line 7:
```suggestion
import { getIPFSHash } from '../ipfs';
import { queryLIKETransactionInfo } from '../like';
import { ArweaveFile } from './types';
import { estimateARPrices, convertARPricesToLIKE, uploadFilesToArweave } from '.';
```

PR #72, opened 2021-10-15
🔧 Add ibc flag in keplr config

williamchong:


store/keplr.ts
```diff
@@ -75,10 +74,14 @@ export default class Keplr extends VuexModule {
           feeCurrencies: network.coinLookup.map(({ viewDenom }) => configToKeplrCoin(viewDenom)),
           coinType: 118,
           gasPriceStep: {
-            low: DEFAULT_GAS_PRICE_NUMBER,
-            average: DEFAULT_GAS_PRICE_NUMBER,
-            high: DEFAULT_GAS_PRICE_NUMBER,
+            low: 0.01,
+            average: 1,
+            high: 1000,
```
nwingt, line 79: Why not use constant here?

nnkken: I think it is too low since we are using `nano` as unit...
Maybe like this?
`{ low: 1, average: 10, high: 1000 }`

williamchong: low must be `0.01` so that ppl without LIKE can register as user (hack since keplr cannot sign arbitrary message, so we are signing a empty transaction with memo, which gas: 1

PR #73, opened 2021-10-17
♻️ Improve the display content in EXIF

AuroraHuang22:


components/MetadataCard.vue
```diff
@@ -23,9 +23,9 @@
     </template>
     <FormField
       v-for="(item, name) in exifInfo"
-      :key="isFromData ? $t(`iscn.exif.label.${name}`) : name"
+      :key="isFromFilterEXif ? $t(`iscn.exif.label.${name}`) : name"
```
nwingt, line 26: 可能 EXIF 統一下大小寫，我建議是
```suggestion
      :key="isFromFilterEXIF ? $t(`iscn.exif.label.${name}`) : name"
```

AuroraHuang22: 🤣 I don't know why I did that

PR #74, opened 2021-10-17
✨ Add the testnet indicator

AuroraHuang22:
截圖 2021-10-18 上午11 24 21


components/AppHeader.vue
```diff
@@ -80,6 +80,13 @@
           </Button>
         </div>
       </nav>
+      <Label
+        :v-if="IS_TESTNET"
```
nwingt, line 84: Are you sure IS_TESTNET is not undefined?

AuroraHuang22: It's a string 'TRUE' in the test environment

PR #75, opened 2021-10-19
♻️ Refactor registration steps

AuroraHuang22:
[Task here](https://www.notion.so/likecoin/Implement-new-registration-process-9d0b9d9dd93842398531d8b063315671)


locales/en.json
```diff
@@ -109,6 +109,10 @@
   "IscnUploaded.button.new": "Register Another ISCN",
   "IscnUploaded.guide.title": "Completed! Here is your ISCN",
   "IscnUploaded.title": "ISCN Registed",
+  "Registration.step.1":"Step 1/4",
+  "Registration.step.2":"Step 2/4",
+  "Registration.step.3":"Step 3/4",
+  "Registration.step.4":"Step 4/4",
```
nwingt, line 115: I suggest to use interpolation for `x/x`

nwingt (commented):
I suggest prefixing `UploadForm` with `Iscn`

nwingt (commented):
I think it is better refactoring the form header, including the step to the parent

components/IscnFormHeader.vue
```diff
@@ -0,0 +1,34 @@
...
+    >
+      <template #prepend>
+        <IconRegister />
+      </template>
+    </Label>
+    <div class="flex flex-col items-end">
+      <Stepper :step="step" />
+      <Label
+        preset="p6"
+        :text="`${$t('Registration.step')} ${step}/4`"
```
nwingt, line 20: Use interpolation is better
```suggestion
        :text="$t('Registration.step', { step, total: 4 })"
```

PR #79, opened 2021-10-29
🐛 Fix timeout issue on large file uploads

AuroraHuang22:
1. Add a warning to prevent users from uploading files exceeding 20 MB.
2. Update the UI of submitting to arweave.
3. Add the timeout config.

### DIFF
--- a/components/IscnUploadForm.vue
+++ b/components/IscnUploadForm.vue
@@ -149,7 +149,6 @@
       v-model="isSizeExceeded"
       :text="$t('UploadForm.warning')"
       preset="warn"
-      :timeout="2000"
     />
   </div>
 </template>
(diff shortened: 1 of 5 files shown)


components/IscnUploadForm.vue
```diff
@@ -173,9 +179,10 @@ export default class UploadForm extends Vue {
   fileName: string = ''
   fileSize: number = 0
   fileType: string = ''
-  uploadStatus: string = '';
+  uploadStatus: string = ''
 
   isOpenFileInfoDialog = false
+  isTooLargeFile = false
```
nwingt, line 185: Can u try using `uploadStatus` with 'size-exceeded' instead of adding a new boolean

AuroraHuang22: I think it would be better to use boolean because we have to bind it with the v-model, also I found that `uploadStatus` is useless in this component, so I decided to remove `uploadStatus`.

nwingt (approved):

PR #80, opened 2021-11-02
♻️ Improve the cancellation behavior

AuroraHuang22:
1. Improved the cancellation behavior for both author editing and signing.
2. Implemented a new signing UI.
[task here](https://www.notion.so/likecoin/Implement-new-signing-UI-d36eb9eac04b4822918916909b99b3bc)

### DIFF
--- a/components/Dialog.vue
+++ b/components/Dialog.vue
@@ -13,7 +13,7 @@
           'backdrop-filter',
           'backdrop-blur-lg',
         ]"
-        @click="handleClickToClose"
+        @click="handleBackdropClick"
       />
       <div
         :class="[
@@ -91,8 +91,8 @@ export default class Dialog extends Vue {
   // Content text.
   @Prop({ default: '' }) readonly text!: string
 
-  // Close the dialog by clicking outside
-  @Prop({ default: true }) readonly isClickToClose!: boolean
+  // Set to true to disable close by backdrop click
+  @Prop({ default: false }) readonly isDisabledBackdropClick!: boolean
 
   // Set to false to hide the close button
   @Prop({ default: true }) readonly hasCloseButton!: boolean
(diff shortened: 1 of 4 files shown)


components/Dialog.vue
```diff
@@ -120,5 +127,12 @@ export default class Dialog extends Vue {
     this.$emit('close')
     this.isOpen = false
   }
+
+  handleClickToClose() {
```
nwingt, line 131: can replace 133 & 134 with `close()`

components/Dialog.vue
```diff
@@ -90,6 +91,12 @@ export default class Dialog extends Vue {
   // Content text.
   @Prop({ default: '' }) readonly text!: string
 
+  // Close the dialog by clicking outside
+  @Prop({ default: true }) readonly isClickToClose!: boolean
```
nwingt, line 95: Set to true to disable close by backdrop click
```suggestion
  @Prop({ default: true }) readonly isDisabledBackdropClick: boolean
```

components/Dialog.vue
```diff
@@ -13,7 +13,7 @@
           'backdrop-filter',
           'backdrop-blur-lg',
         ]"
-        @click="close"
+        @click="handleClickToClose"
```
nwingt, line 16:
```suggestion
        @click="handleBackdropClick"
```

nwingt (changes requested):

components/Dialog.vue
```diff
@@ -90,6 +91,12 @@ export default class Dialog extends Vue {
   // Content text.
   @Prop({ default: '' }) readonly text!: string
 
+  // Set to true to disable close by backdrop click
+  @Prop({ default: true }) readonly isDisabledBackdropClick!: boolean
```
nwingt, line 95: Should be default `false`?

nwingt (approved):

nwingt (approved):

PR #81, opened 2021-11-08
Hotfix

AuroraHuang22:


nwingt (changes requested):
The issue happens because of the extra `:` before `v-if`

https://github.com/likecoin/app-like-co/blob/3402b2d3eb9bec35846d94238c0e2d66230f92a2/components/AppHeader.vue#L84

PR #83, opened 2021-11-12
✨ Enable users to get token

AuroraHuang22:
[task 1](https://www.notion.so/likecoin/app-like-co-Add-get-token-pills-on-index-60b42c28bbf44d98b725822f88e5b551) & [task 2](https://www.notion.so/likecoin/app-like-co-Redirect-to-osmosis-if-the-insufficient-balance-d34623e0080941fc90b9a5197c9edbcd)


pages/index.vue
```diff
@@ -59,6 +60,41 @@
...
+        href="https://app.osmosis.zone/"
+        :text="$t('HomePage.label.osmosis')"
+      />
+      <Button
+        class="mr-[24px]"
+        preset="tertiary"
+        text-preset="h6"
+        size="mini"
+        href="https://app.liquid.com/exchange/LIKEUSDT"
+        :text="$t('HomePage.label.liquid')"
```
nwingt, line 95: The translation key should be `button`? Instead of `label`

nwingt (approved):

PR #89, opened 2021-12-03
✨ Support login with WalletConnect

nwingt:


williamchong (approved):
Strange that the store is called `keplr` but has walletconnect codes, maybe rename to wallet?

nwingt:
@williamchong Renamed

constant/index.ts
```diff
@@ -54,3 +54,8 @@ export const WALLET_TYPES = [
 ]
 
 export const SITE_URL = IS_TESTNET ? 'https://app.rinkeby.like.co' : 'https://app.like.co';
+
+export const CONNECT_WALLET_TYPES = [
+  'keplr',
+  'likerId',
```
williamchong, line 60: should this be named `likerland` as in `likerland app`

williamchong (changes requested):

constant/index.ts
nwingt: How about `likerland_app`?

williamchong: sure

williamchong (approved):

PR #90, opened 2021-12-09
✨ Add multi ipfs server support

williamchong:
Allow making copy to our internal ipfs server


nnkken (approved):

williamchong:
@nwingt Do you think this should be merged before or after airdrop PRs? Should be non-breaking

nwingt (approved):

PR #102, opened 2021-12-22
✨ Check format of wallet address

AuroraHuang22:


components/AirdropLogin.vue
```diff
@@ -108,22 +109,46 @@
...
 
+export enum CosmosArrdessFormat {
+  head = 'cosmos1',
+  length = 45
+}
+export enum OsmoArrdessFormat {
+  head = 'osmo1',
+  length = 43
+}
+
```
williamchong, line 123: I think using regex is cleaner, and can filter invalid characters at the same time

williamchong (changes requested):

components/AirdropLogin.vue
```diff
@@ -108,22 +109,48 @@
...
+} from '~/constant'
+
+export enum CosmosArrdessFormat {
+  head = 'cosmos1',
+  length = 45
+}
+export enum OsmoArrdessFormat {
+  head = 'osmo1',
+  length = 43
+}
```
williamchong, line 126: Remove if unused

williamchong (changes requested):

components/AirdropLogin.vue
```diff
@@ -108,22 +109,39 @@
...
   handleAddressInput() {
-    this.$emit('input', this.inputAddress)
+    this.errorMessage = ''
+    if (
+      COSMOS_ADDRESS_REGEX.test(this.inputAddress) ||
+      OSMO_ADDRESS_REGEX.test(this.inputAddress)
+    ) {
+      this.$emit('input', this.inputAddress)
+    }
+    this.errorMessage = this.$t('AirDrop.errorMessage.address') as string
```
williamchong, line 144: Missing return, even if this passes the test, `this.errorMessage` would still be set
Also prefer guarding for error using i, and emit normal case at the end of function, rather than checking for valid case using if, and throwing error as default

components/AirdropLogin.vue
```diff
@@ -108,22 +109,39 @@
...
   // eslint-disable-next-line class-methods-use-this
   get connectWalletTypes() {
     return CONNECT_WALLET_TYPES
   }
 
+  @Watch('inputAddress')
+  onValueChange() {
+    this.errorMessage = ''
+  }
+
```
williamchong, line 135: I think you can use `@input` on `textfield` instead of using watch, avoid watch if possible

williamchong (changes requested):

components/AirdropLogin.vue
```diff
@@ -109,21 +110,34 @@
...
   handleAddressInput() {
+    this.errorMessage = ''
+    if (
+      !COSMOS_ADDRESS_REGEX.test(this.inputAddress) ||
+      !OSMO_ADDRESS_REGEX.test(this.inputAddress)
+    ) {
+      this.errorMessage = this.$t('AirDrop.errorMessage.address') as string
+    }
     this.$emit('input', this.inputAddress)
+    this.errorMessage = ''
```
williamchong, line 140:
```suggestion
    this.errorMessage = ''
    if (
      !COSMOS_ADDRESS_REGEX.test(this.inputAddress) ||
      !OSMO_ADDRESS_REGEX.test(this.inputAddress)
    ) {
      this.errorMessage = this.$t('AirDrop.errorMessage.address') as string
      return;
    }
    this.$emit('input', this.inputAddress)
```

williamchong: `this.errorMessage` will never be shown in your version if you always empty it?

williamchong (changes requested):

components/AirdropLogin.vue
AuroraHuang22: soga

constant/index.ts
```diff
@@ -63,6 +63,10 @@ export const CONNECT_WALLET_TYPES = [
 
 export const EMAIL_REGEX = /(\S)+[@]{1}(\S)+[.]{1}(\w)+/
 
+export const COSMOS_ADDRESS_REGEX = /^cosmos1[ac-hj-np-z02-9]{38}$/
+
+export const OSMO_ADDRESS_REGEX = /^osmo1[ac-hj-np-z02-9]{38}$/
```
nwingt, line 68:
```suggestion
export const OSMOSIS_ADDRESS_REGEX = /^osmo1[ac-hj-np-z02-9]{38}$/
```

components/AirdropLogin.vue
```diff
@@ -109,20 +110,33 @@
 </template>
 <script lang="ts">
 import { Vue, Component, Prop } from 'vue-property-decorator'
-import { CONNECT_WALLET_TYPES } from '~/constant'
+import {
+  CONNECT_WALLET_TYPES,
+  COSMOS_ADDRESS_REGEX,
+  OSMO_ADDRESS_REGEX,
```
nwingt, line 116:
```suggestion
  OSMOSIS_ADDRESS_REGEX,
```

components/AirdropLogin.vue
```diff
@@ -109,20 +110,33 @@
...
   // eslint-disable-next-line class-methods-use-this
   get connectWalletTypes() {
     return CONNECT_WALLET_TYPES
   }
 
   handleAddressInput() {
+    this.errorMessage = ''
+    if (
+      !COSMOS_ADDRESS_REGEX.test(this.inputAddress) ||
+      !OSMO_ADDRESS_REGEX.test(this.inputAddress)
```
nwingt, line 135:
```suggestion
      !OSMOSIS_ADDRESS_REGEX.test(this.inputAddress)
```

nwingt (changes requested):

nwingt (approved):

PR #118, opened 2022-01-19
Add pagination in /works

nwingt:
image


pages/works/index.vue
```diff
@@ -42,10 +119,30 @@ export default class WorksIndexPageextends extends Vue {
 
   async refreshWorks() {
     if (!this.currentAddress) {
-      this.records = []
-      return
+      this.pages = []
+    } else {
+      const records = await this.queryISCNByAddress(this.currentAddress)
```
williamchong, line 124: Not sure if we want to also handle nextSequence in the api level
ref: https://github.com/likecoin/iscn-js/blob/master/src/queryClient.ts#L101

williamchong (approved):

PR #127, opened 2022-03-23
Integrate Numbers Protocol

nwingt:


components/CheckBox.vue
```diff
@@ -0,0 +1,61 @@
...
+      <IconCheckBoxOff
+        v-else
+        :class="iconClass"
+      />
+      <slot />
+    </label>
+  </span>
+</template>
+
+<script>
```
AuroraHuang22, line 35: Not using TypeScript?

server/arweave/router.ts
```diff
@@ -81,9 +59,17 @@ router.post('/upload',
 
     // shortcut for existing file without checking tx
     if (existingArweaveId) {
+      let numAssetsIds = [];
+      if (req.body.num === '1') {
```
williamchong, line 63: Would `req.body.num && req.body.num !== '0'` be more robust?

components/CheckBox.vue
williamchong, line 35:
```suggestion
<script lang="ts">
```

williamchong (changes requested):

components/CheckBox.vue
nwingt: Updated

server/arweave/router.ts
```diff
@@ -115,11 +101,25 @@ router.post('/upload',
...
       return;
     }
-    const [{ arweaveId, list }] = await Promise.all([
+    const promises: Promise<any>[] = [
       uploadFilesToArweave(arFiles),
       uploadFilesToIPFS(arFiles),
-    ]);
-    res.json({ arweaveId, ipfsHash, list });
+    ];
+    if (req.body.num === '1') {
```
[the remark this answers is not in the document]
nwingt: Updated

server/arweave/router.ts
nwingt: Updated

components/CheckBox.vue
nwingt: Updated

williamchong (approved):

PR #131, opened 2022-03-29
💄 UI improvement in [View/details] page

AuroraHuang22:
- correct the version by fetching data from records data instead of metadata
- show file preview as long as the content type is image
- add progress indicator for loading preview image

### DIFF
--- a/components/MetadataCard.vue
+++ b/components/MetadataCard.vue
@@ -14,12 +14,12 @@
       </template>
     </Label>
     <template v-if="imgSrc">
-      <ProgressIndicator v-if="isLoading" class="mt-[24px]"/>
+      <ProgressIndicator v-if="!hasLoadedImage" class="mt-[24px]"/>
       <img
         ref="iscnImg"
         class="block w-full mt-[24px] rounded-[8px]"
         :src="imgSrc"
-        @load="isLoading = false"
+        @load="hasLoadedImage = true"
       />
       <FormField class="mt-[8px]" :label="$t('iscn.meta.card.preview')" />
     </template>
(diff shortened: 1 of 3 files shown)


pages/view/_iscnId.vue
```diff
@@ -467,7 +467,7 @@ export default class ViewIscnIdPage extends Vue {
   }
 
   get recordData() {
-    return this.record?.data
+    return this.record && this.record.data
```
nwingt, line 470: TypeScript not working?

pages/view/_iscnId.vue
```diff
@@ -505,6 +505,10 @@ export default class ViewIscnIdPage extends Vue {
     return `${ISCN_RAW_DATA_ENDPOINT}${this.iscnId}`
   }
 
+  get version() {
+    return this.recordData.recordVersion
```
nwingt, line 508: `recordData` may be undefined,

components/MetadataCard.vue
```diff
@@ -57,6 +59,7 @@ export default class MetadataCard extends Vue {
   @Prop(Object) readonly allExif!: Object
 
   exifInfo: Object = {}
+  isLoading: boolean = true
```
nwingt, line 62: I suggest renaming this to more specify `hasLoadedImage`, with default is `false`

nwingt (changes requested):

pages/view/_iscnId.vue
AuroraHuang22: After I re-run dev it works fine.

williamchong (approved):

PR #135, opened 2022-04-07
🎨 Replace legacy LCD endpoints

WeiJun0827:


constant/index.ts
```diff
@@ -32,8 +32,8 @@ export const RAWDATA_URL = {
 }
 
 export const RAWDATA_TX_URL = {
-  testnet: 'https://node.testnet.like.co/txs?iscn_record.iscn_id=',
-  production: 'https://mainnet-node.like.co/txs?iscn_record.iscn_id=',
+  testnet: 'https://node.testnet.like.co/cosmos/tx/v1beta1/txs?events=iscn_record.iscn_id=',
+  production: 'https://mainnet-node.like.co/cosmos/tx/v1beta1/txs?events=iscn_record.iscn_id=',
```
williamchong, line 36: Seems this query actually only wants `create_iscn_record` event, can we also filter it here ? https://github.com/likecoin/app-like-co/blob/41eae435240d6f86006bdc7437338854f5734eed/pages/view/_iscnId.vue#L623

williamchong (changes requested):

williamchong (approved):

PR #137, opened 2022-04-15
🐛 Fix missing numAssetIds

AuroraHuang22:


server/arweave/router.ts
```diff
@@ -117,7 +117,6 @@ router.post('/upload',
         arweaveId,
         list,
       },
-      ,
       numAssetIds,
     ] = await Promise.all(promises)
```
williamchong, line 121: promises in order would be `uploadFilesToArweave`, and `uploadFilesToIPFS` and possibly `registerNUMAssets(` in the third element, so that comma should be needed and not an error?

components/IscnRegisterForm.vue
```diff
@@ -1028,7 +1028,7 @@ export default class IscnRegisterForm extends Vue {
     try {
       const {
         arweaveId,
-        numAssetIds: [numbersProtocolAssetId],
+        numAssetIds: {numbersProtocolAssetId},
```
williamchong, line 1031: But it seems that from the API, `numAssetIds` is really an array of assetId?

williamchong (changes requested):
Not really sure why some of the change is made

server/arweave/router.ts
williamchong: I think adding `_` to indicate unused param for the 2nd promise is a good idea though

constant/index.ts
```diff
@@ -32,7 +32,7 @@ export const RAWDATA_URL = {
 }
 
 export const RAWDATA_TX_URL = {
-  testnet: 'https://node.testnet.like.co/cosmos/tx/v1beta1/txs?events=message.action=\'create_iscn_record\'&events=iscn_record.iscn_id=',
+  testnet: 'https://node.testnet.like.co/cosmos/tx/v1beta1/txs?events=message.action=\'/likechain.iscn.MsgCreateIscnRecord\'&events=iscn_record.iscn_id=',
```
WeiJun0827, line 35: Should we keep both URLs to support backward compatibility?

constant/network.ts
```diff
@@ -25,6 +25,7 @@ const testnetConfig = {
   name: 'LikeCoin Taipei test chain',
```
WeiJun0827, line 25: Is `LikeCoin public test chain` preferred?

constant/index.ts
AuroraHuang22: I think we don't need to support backward in the testnet,(since there is no old data in the public-testnet-5🤔), but it will be needed in the mainnet.

server/arweave/router.ts
```diff
@@ -70,7 +70,7 @@ router.post('/upload',
       res.json({
         arweaveId: existingArweaveId,
         ipfsHash,
-        numAssetsIds,
+        numAssetsIds: numAssetsIds || [],
```
williamchong, line 73: I wonder in what case `await registerNUMAssets` would return empty array though...

williamchong (approved):

server/arweave/router.ts
AuroraHuang22: Oops, my bad 😓

PR #138, opened 2022-04-18
💬 Update error message for not eligible for the airdrop

AuroraHuang22:
截圖 2022-04-18 下午9 54 01


williamchong (approved):

williamchong:
I think `v-if="errorMessage === $t('AirDrop.errorMessage.ineligible')"` is kind of a poor condition though, would be nice if we can use `'AirDrop.errorMessage.ineligible'` directly for comparison somehow

nwingt (approved):

PR #142, opened 2022-05-03
💄 [/RootLayout] Add blocking message for chain upgrade

AuroraHuang22:
截圖 2022-05-03 下午6 37 28

### DIFF
--- a/components/RootLayout.vue
+++ b/components/RootLayout.vue
@@ -12,10 +12,10 @@
 
     <DialogContainer>
       <Dialog
-        :open="isOpenBlockingDialog"
+        :open="isOpenChainUpgradeBlockingDialog"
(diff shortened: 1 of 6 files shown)


williamchong (commented):
can we block any "submit" button that is related to chain also?

components/RootLayout.vue
```diff
@@ -9,6 +9,37 @@
     ]"
   >
     <slot />
+
+    <DialogContainer>
+      <Dialog
+        :open="isOpenBlockingDialog"
```
williamchong, line 15: `isOpenChainUpgradeBlockingDialog` in case we have other blocking dialog

williamchong (approved):

williamchong (approved):

PR #145, opened 2022-05-06
🐛 Fix ISCN missing issue when ISCN number > 100

WeiJun0827:


store/iscn.ts
```diff
@@ -19,23 +20,28 @@ export default class ISCN extends VuexModule {
...
 
   @Mutation
   setRecords(records: ISCNRecord[]) {
     records.forEach((r) => {
       const id = r.data['@id'] as string;
       Vue.set(this.recordsById, id, { id, ...r })
-      Vue.set(this.recordsByIPLD, r.ipld, { id, ...r })
+      this.records.push({ id, ...r })
     })
   }
```
williamchong, line 34: Since this function involves `.push()` now I don't think `setRecords` is a good name, maybe `addRecords` or `appendRecords` is better

williamchong (changes requested):

williamchong (approved):

PR #147, opened 2022-05-13
🚸  Access search parameter in url

AuroraHuang22:


components/WalletFieldList.vue
```diff
@@ -117,6 +118,11 @@ export default class WalletFieldList extends Vue {
       isOpenOptions: false,
     }]
 
+  // eslint-disable-next-line class-methods-use-this
+  mounted(){
+    console.log(this.value)
+  }
+
```
nwingt, line 125: To be removed?

pages/search/_keyword.vue
```diff
@@ -0,0 +1,152 @@
...
+
+  get pages() {
+    return this.recordChunks || []
+  }
+
+  async mounted() {
+    if (!this.keyword) {
+      this.$router.push(this.localeLocation({ name: 'index' })!)
+      return
+    }
```
nwingt, line 137: It is more robust to guard in `fetch()` or place it in `index.vue`

nwingt:
I suggest adding a subtitle for search result to distinguish from portfolio page

image

williamchong (changes requested):

store/iscn.ts
```diff
@@ -86,57 +86,74 @@ export default class ISCN extends VuexModule {
...
+    try {
+      const client = await getQueryClient()
+      let res
+      let nextSequence
+      do {
+        this.context.commit('setIsLoading', true)
+        // eslint-disable-next-line no-await-in-loop
+        res = await client
+          .queryRecordsByOwner(address, nextSequence)
+          .catch(() => {})
```
williamchong, line 99: why add extra catch here, and not let the error get catch below?

store/iscn.ts
```diff
@@ -86,57 +86,74 @@ export default class ISCN extends VuexModule {
...
+        if (res) {
+          nextSequence = res.nextSequence.toNumber()
+          this.context.commit('appendRecords', res.records)
+          records = records.concat(res.records)
+        }
+      } while (nextSequence !== 0)
+    } catch (error) {
+      console.error(error)
```
williamchong, line 107: Do we want to throw the error for caller to handle error instead?

store/iscn.ts
```diff
@@ -86,57 +86,74 @@ export default class ISCN extends VuexModule {
...
+        ).filter((t) => t) as ISCNRecord[]
+      }
+      result = result
+        .concat(txRecords)
+        .concat(idRes ? idRes.records : [])
+        .concat(fingerprintRes ? fingerprintRes.records : [])
+        .concat(ownerRes ? ownerRes.records : [])
+      this.context.commit('appendRecords', result)
+    } catch (error) {
+      console.error(error)
```
williamchong, line 153: Similar to above, is catching the error here intended?

store/iscn.ts
AuroraHuang22: Add an error handler to open a warning snackbar in page.

store/iscn.ts
AuroraHuang22: I think each request in [Promise.all] should have its own catch, since there might be an error if the keyword format is not suitable for some APIs, I keep another catch outside to handle errors other than above.

PR #152, opened 2022-06-14
✨ Add NFT portal

AuroraHuang22:


pages/nfttest/mint/_iscnId.vue
```diff
@@ -74,28 +178,26 @@ export default class NFTTestMintPage extends Vue {
...
-      case 'mint':
-        await this.sendNFT();
-        await this.postMintInfo();
-        await this.getMintInfo();
-        break;
+        this.isLoading = true
+        await this.createNftClass()
+        await this.sendNFT()
+        await this.postMintInfo()
+        await this.getMintInfo()
```
williamchong, line 187: if you catch the error in each function, you would have to check return value of each function, otherwise it will continue on and send wrong payload in the next step

pages/nfttest/mint/_iscnId.vue
```diff
@@ -105,73 +207,81 @@ export default class NFTTestMintPage extends Vue {
...
+            iscn_id: this.iscnId,
+            class_id: this.classId,
+          },
+          paramsSerializer: (params) => qs.stringify(params),
+        },
+      )
+      fdata = data
+    } catch (error) {
+      console.error(error)
+    }
```
williamchong, line 236: I think throwing the error might allow some simple error display on UI? Or we can also set a global error data to display the error on UI

pages/nfttest/mint/_iscnId.vue
```diff
@@ -74,28 +178,26 @@ export default class NFTTestMintPage extends Vue {
...
     /* eslint-enable no-fallthrough */
   }
 
   async getISCNInfo() {
-    const res = await this.fetchISCNById(this.iscnId);
+    const res = await this.fetchISCNById(this.iscnId)
     if (res) {
-      console.log(res);
-      [this.iscnData] = res.records;
+      ;[this.iscnData] = res.records
```
williamchong, line 200: Not sure if it should be `this.iscnData = res.records[0].data` instead, please check

williamchong (changes requested):

pages/nfttest/mint/_iscnId.vue
```diff
@@ -172,18 +181,35 @@ export default class NFTTestMintPage extends Vue {
   }
 
   async mounted() {
-    await this.getMintInfo()
+    this.getMintInfo()
     await this.getISCNInfo()
```
williamchong, line 185: Please use Promise.all if they are all async but we don't need them to be blocking

PR #154, opened 2022-06-15
🐛 Fix keplr detection condition

williamchong:


store/wallet.ts
```diff
@@ -97,13 +97,14 @@ export default class Wallet extends VuexModule {
...
       const TRY_COUNT = 3;
-      while (TRY_COUNT > tries) {
+      while (!window.keplr && TRY_COUNT > tries) {
         // eslint-disable-next-line no-await-in-loop
         await timeout(1000);
         tries += 1;
       }
     }
     if (window.keplr && window.keplr.experimentalSuggestChain) {
+      this.toggleKeplrWarning(false)
```
AuroraHuang22, line 107: my suggestion :
```suggestion
    if (this.type === 'keplr') return true;
    let tries = 0;
    const TRY_COUNT = 3;
    while (!window.keplr && TRY_COUNT > tries) {
      // eslint-disable-next-line no-await-in-loop
      await timeout(1000);
      tries += 1;
    }
    if (!window.keplr || !window.keplr.experimentalSuggestChain) {
      this.toggleKeplrWarning(true);
      return false;
    }
    try { ......
```
```

williamchong: Would this break if user keplr is locked/removed after logging in once with keplr?

williamchong:
replaced by https://github.com/likecoin/app-like-co/pull/155

PR #155, opened 2022-06-21
🐛 Fix keplr detection condition

AuroraHuang22:


williamchong (changes requested):
The error would show up 3seconds after I click keplr button right? Can we make it shorter.
I think reducing `TRY_COUNT` is ok too

williamchong (approved):

PR #162, opened 2022-07-23
✨ Upload formatted HTML body and OG image to Arweave, and store ID in metadata

WeiJun0827:


server/crawler/index.ts
```diff
@@ -1,28 +1,209 @@
...
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>${title}</title>
+    <meta name="description" content="${description}">
+    <meta property="og:title" content="${author}: ${title}">
+    <meta property="og:description" content="${description}">
+    <meta property="article:author" content="${author}">
+    <meta name="twitter:title" content="${author}: ${title}">
+    <meta name="twitter:description" content="${description}">
```
williamchong, line 158: Can we not set the relevant og if the field does not exists at all?

server/crawler/index.ts
```diff
@@ -1,28 +1,209 @@
...
+  title,
+  author,
+  description,
+}: {
+  content: string,
+  title: string,
+  author: string,
+  description: string,
+}) {
+  const style = `
```
williamchong, line 15: Please add a comment to credit matters-html-formatter

pages/nfttest/mint/_iscnId.vue
```diff
@@ -272,6 +274,79 @@ export default class NFTTestMintPage extends Vue {
     }
   }
 
+  async getOgImage() {
+    try {
+      const { url } = this.iscnData.contentMetadata
+      if (!url) { return }
```
williamchong, line 280: Please guard `contentMetadata` does not exists too

pages/nfttest/mint/_iscnId.vue
```diff
@@ -272,6 +274,79 @@ export default class NFTTestMintPage extends Vue {
...
+      const { data: { image } } = await this.$axios.get(`/crawler/?url=${encodeURIComponent(url)}`)
```
williamchong, line 281: Can we pass this information from last page if they came from the fetch page?
We can fetch the API again if this info does not exists (ie did not came from fetch page)

pages/nfttest/mint/_iscnId.vue
```diff
@@ -191,14 +177,21 @@ export default class NFTTestMintPage extends Vue {
   get loadingText(): string {
     if (this.state === 'done') return ''
     if (this.state === 'mint') return 'Minting NFT ...'
+    if (this.ogImageFormData && !this.ogImageArweaveId) return 'Uploading display image ...'
     return 'Creating NFT class ...'
   }
 
+  get ogImageArweaveIdWithPrefix(): string {
```
williamchong, line 184: `ogImageArweaveUri`

pages/nfttest/mint/_iscnId.vue
```diff
@@ -272,6 +274,79 @@ export default class NFTTestMintPage extends Vue {
...
+    try {
+      const { url } = this.iscnData.contentMetadata
+      if (!url) { return }
+      const { data: { image } } = await this.$axios.get(`/crawler/?url=${encodeURIComponent(url)}`)
+      if (!image) { return }
+      const { data, headers } = await this.$axios.get(image)
+      const blob = new Blob([data], { type: headers['content-type'] })
+      const formData = new FormData()
+      formData.append('file', blob)
+      this.ogImageFormData = formData
```
williamchong, line 287: Storing image blob as `data`, and formdata as `computed` make this more readable

pages/nfttest/mint/_iscnId.vue
```diff
@@ -272,6 +274,79 @@ export default class NFTTestMintPage extends Vue {
...
+      // eslint-disable-next-line no-console
+      console.error('CANNOT_SEND_ARWEAVE_FEE_TX')
+      throw err
+    }
+  }
+
+  async submitToArweave(): Promise<void> {
+    try {
+      const { arweaveId } = await this.$axios.$post(
+        `${API_POST_ARWEAVE_UPLOAD}?txHash=${this.ogImageTxHash}`,
```
williamchong, line 334: I think `ogImageTxHash` can be passed as function param instead of this.data here, you don't use the hash any where else.

pages/nfttest/fetch/index.vue
```diff
@@ -188,15 +219,82 @@ export default class FetchIndex extends Vue {
...
+      // eslint-disable-next-line no-console
+      console.error('CANNOT_SEND_ARWEAVE_FEE_TX')
+      throw err
+    }
+  }
+
+  async submitToArweave(): Promise<void> {
+    try {
+      const { arweaveId } = await this.$axios.$post(
+        `${API_POST_ARWEAVE_UPLOAD}?txHash=${this.txHash}`,
```
williamchong, line 274: I think `txHash ` can be passed as function param instead of this.data here, you don't use the hash any where else.
Also the name `txHash` is confusing

pages/nfttest/fetch/index.vue
```diff
@@ -102,11 +106,52 @@ export default class FetchIndex extends Vue {
...
   ownerWallet = ''
   errorMessage = ''
-  state = 'init'
+  crawledData: any
+  txHash = ''
+  ipfsHash = ''
+  arweaveId = ''
+  arweaveFeeTargetAddress = ''
+  arweaveFee = new BigNumber(0)
+  memo = ''
```
williamchong, line 115: Where are these used?

pages/nfttest/fetch/index.vue
```diff
@@ -102,11 +106,52 @@ export default class FetchIndex extends Vue {
...
+    return {
+      type: 'CreativeWork',
+      name: title,
+      description,
+      tagsString: keywords,
+      url: this.url,
+      exifInfo: {},
+      license: '',
+      ipfsHash: this.ipfsHash,
+      arweaveId: this.arweaveId,
```
williamchong, line 141: Do we prefer `undefined` here if they are not set?

pages/nfttest/fetch/index.vue
```diff
@@ -102,11 +106,52 @@ export default class FetchIndex extends Vue {
...
+      fileSHA256: '',
+      authorNames: ['Author'],
```
williamchong, line 143: Why `Author`?

pages/nfttest/fetch/index.vue
```diff
@@ -102,11 +106,52 @@ export default class FetchIndex extends Vue {
...
+      authorUrls: [['']],
+      authorWallets: [[{
+        address: this.address,
+        type: 'like',
+      }]],
+      likerIds: [''],
+      descriptions: [''],
+      numbersProtocolAssetId: '',
```
williamchong, line 151: Don't set to empty if not exists

pages/nfttest/fetch/index.vue
```diff
@@ -142,42 +187,28 @@ export default class FetchIndex extends Vue {
...
-      descriptions: [''],
-      numbersProtocolAssetId: '',
+      this.isLoading = true
+      await this.crawlUrlData()
+      const arweaveFeeInfo = await this.estimateArweaveFee()
+      if (!this.arweaveId) {
+        await this.sendArweaveFeeTx(arweaveFeeInfo)
+        await this.submitToArweave()
+      }
+      await this.submitToISCN()
```
williamchong, line 205: What happen if `crawlUrlData` cannot fetch anything?

williamchong (changes requested):

pages/nfttest/fetch/index.vue
WeiJun0827: hmmm I think using `if` to guard it should be more clear

williamchong: return undefined would work if they are computed

server/crawler/index.ts
```diff
@@ -12,136 +13,149 @@ function formatBody({
...
   }
-  blockquote {
-    margin-left: 20px;
-    margin-right: 20px;
-    color: #5F5F5F;
+  if (author) {
+    meta += `
+    <meta property="article:author" content="${author}">
+    <meta property="og:title" content="${author}: ${title}">
+    <meta name="twitter:title" content="${author}: ${title}">`
```
williamchong, line 27: If author not exists, title should just be `title`, without `og:title` we will have many preview display issue

pages/nfttest/fetch/index.vue
```diff
@@ -307,6 +322,7 @@ export default class FetchIndex extends Vue {
           this.localeLocation({
             name: 'nfttest-mint-iscnId',
             params: { iscnId: this.iscnId },
+            query: this.mintQuery,
```
williamchong, line 298: Any reason why not use `params` instead of `query`?

pages/nfttest/fetch/index.vue
```diff
@@ -155,6 +155,14 @@ export default class FetchIndex extends Vue {
     }
   }
 
+  get mintQuery() {
+    const query: any = {}
+    if (this.crawledData?.image) {
+      query.ogImageUrl = encodeURIComponent(this.crawledData.image)
```
williamchong, line 134: I think you don't need to call `encodeURIComponent` for query, vue-router would handle that

pages/nfttest/mint/_iscnId.vue
```diff
@@ -118,6 +118,7 @@ export default class NFTTestMintPage extends Vue {
   iscnOwner: string = ''
   iscnData: any = null
   apiData: any = null
+  ogImageUrl = decodeURIComponent(this.$route.query.ogImageUrl as string) || ''
```
williamchong, line 121: You don't need to call `decodeURIComponent ` for query if you just passed it plainly, vue-router would handle that

williamchong (changes requested):

pages/nfttest/fetch/index.vue
WeiJun0827: Currently, it's `/mint/{iscnId}?ogImageUrl={ogImageUrl}`. Should we make it like `/mint/{iscnId}/{ogImageUrl}`? I think `ogImageUrl` is a passing value for `iscnId`. And we don't want to change the file structure just for accepting `ogImageUrl`?

williamchong: The trick is you dont need a param to be in the path, but you can still pass it as a param as long as it is defined in the page, and passed when the page is called in router

williamchong: i.e. try just putting `params: { iscnId: this.iscnId, ogImageUrl: this.crawledData.image }` and see if `$route.params.ogImageUrl` exists in the next page

WeiJun0827: I see, it works! Thanks!!

pages/nft/iscn/_iscnId.vue
```diff
@@ -188,6 +187,11 @@ export default class NFTTestMintPage extends Vue {
     return 'Creating NFT class ...'
   }
 
+  get ogImageArweaveUri(): string {
+    if (!this.ogImageArweaveId) return ''
+    return `ar://${this.ogImageArweaveId}`
+  }
```
williamchong, line 193: Can we fallback to `getNftClassImage` if `ogImageArweaveId ` not exists
Also I think there is a bug in the original code, we don't know the class id when creating class. In that case, don't set image

williamchong (changes requested):

williamchong (approved):

PR #163, opened 2022-07-25
✨ Add post mapping in fetch iscn page

Tsai-Hsueh-Kuan:
✨ Add post mapping in fetch iscn page


utils/mapping.js
```diff
@@ -0,0 +1,81 @@
...
+
+async function signLogin(signPayload,signer,address) {
+  if (!window.keplr) throw new Error('CANNOT_FIND_KEPLR');
+  window.keplr.defaultOptions = window.keplr.defaultOptions || {};
+  const originalConfig = window.keplr.defaultOptions.sign || {};
+  window.keplr.defaultOptions.sign = {
+    ...originalConfig,
+    preferNoSetFee: true,
+    preferNoSetMemo: true,
+  };
```
williamchong, line 44: Please remove these for now since we might be on wallet connect instead of keplr

utils/mapping.js
```diff
@@ -0,0 +1,81 @@
...
+  const message = {
+    chain_id: network.id,
+    memo: signPayload,
+    msgs: [],
+    fee: { gas: '1', amount: [{ denom: 'nanolike', amount: '0' }] },
```
williamchong, line 49: denom should fetch from `network` also

utils/mapping.js
```diff
@@ -0,0 +1,81 @@
...
+    sequence: '0',
+    account_number: '0',
+  };
+  const payload = await signer.sign(
+    address,
+    message,
+  );
+  window.keplr.defaultOptions.sign = originalConfig;
```
williamchong, line 57: Please remove as above

utils/mapping.js
```diff
@@ -0,0 +1,81 @@
...
+  return { message, ...payload };
+}
+
+async function signByCosmosWallet(source, iscnId, url, likerId, signer, address) {
+ const platform = address.startsWith('like') ? 'likeWallet' : 'cosmosWallet';
```
williamchong, line 62: We don't need to judge `platform` here, please remove

utils/mapping.js
```diff
@@ -0,0 +1,81 @@
...
+    signature,
+    publicKey: publicKey.value,
+    message: stringify(message),
+    from: inputWallet,
+    platform,
+  };
+  return data;
+}
+
+async function signLogin(signPayload,signer,address) {
```
williamchong, line 36: This is not signing login, but arbitary text payload, please rename

utils/mapping.js
```diff
@@ -0,0 +1,81 @@
...
+import { API_LIKER_NFT_MAPPING } from '../constant/api'
+
+async function signISCNMapping(inputWallet, signer, platform, iscnId, url, likerId) {
+  if (!inputWallet) return null;
+  if (![
+'likeWallet',
+'cosmosWallet',
+].includes(platform)) {
+    throw new Error('SIGN_COSMOS_LOGIN_INVALID_PLATFORM');
+  }
```
williamchong, line 13: I think no need to check for these

utils/mapping.js
```diff
@@ -0,0 +1,81 @@
...
+  const ts = Date.now();
+  const payload = JSON.stringify({
+    ts,
+    iscnId,
+    url,
+    likerId,
+    [platform]: inputWallet,
```
williamchong, line 20: we can stick to `likeWallet` and reject anything not `like1`

utils/Keplr.js
```diff
@@ -0,0 +1,144 @@
+/* eslint-disable max-len */
```
williamchong, line 1: Is this file still used?

utils/mapping.js
```diff
@@ -0,0 +1,81 @@
...
+ const platform = address.startsWith('like') ? 'likeWallet' : 'cosmosWallet';
+ const payload = await signISCNMapping(
+   address,
+   s => signLogin(s, signer, address),
+   platform,
+   iscnId,
+   url,
+   likerId,
+ );
+ return { platform, address, payload: { cosmosWalletSource: source, ...payload } };
```
williamchong, line 71: We don't need `cosmosWalletSource` here

williamchong (changes requested):
No need to wrap the signing function by 3 layers, probably 1 or 2 would do

pages/nfttest/fetch/index.vue
```diff
@@ -201,6 +204,7 @@ export default class FetchIndex extends Vue {
         this.address,
       )
       this.iscnId = res.iscnId
+      await postMappingWithCosmosWallet(this.iscnId, this.url, this.likerId, this.signer, this.address)
```
williamchong, line 207: Please only call this when `likerId` and `url` both exists

utils/mapping.js
```diff
@@ -0,0 +1,63 @@
...
+   address,
+   s => payloadSigner(s, signer, address),
+   iscnId,
+   url,
+   likerId,
+ );
+ return { address, payload };
+}
+
+export default async function postWithCosmosWallet(iscnId, url, likerId, signer, address) {
```
williamchong, line 56: Its called `postMappingWithCosmosWallet` in the page, naming issue?

utils/mapping.js
```diff
@@ -0,0 +1,63 @@
...
+ try {
+   const { payload } = await signByCosmosWallet(iscnId, url, likerId, signer, address);
```
williamchong, line 58: We can just call `signISCNMapping ` here without wrapping it in `signByCosmosWallet`

williamchong (changes requested):

williamchong (approved):

PR #175, opened 2022-07-28
:zap: Replace search to indexer

lancatlin:
+ replace search function to call indexer instead of lcd
- concern: tx is not included in indexer search all api

### DIFF
--- a/store/iscn.ts
+++ b/store/iscn.ts
@@ -4,8 +4,10 @@ import { Module, VuexModule, Mutation, Action } from 'vuex-module-decorators';
 import Vue from 'vue';
 import { ISCNRecord } from '@likecoin/iscn-js';
 import _ from 'lodash';
+import axios from 'axios';
 import getQueryClient from '~/utils/cosmos/iscn/query';
 import { ISCNRecordWithID } from '~/utils/cosmos/iscn/iscn.type';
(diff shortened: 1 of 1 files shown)


store/iscn.ts
```diff
@@ -118,40 +120,11 @@ export default class ISCN extends VuexModule {
...
-            }),
-          )
-        ).filter((t) => t) as ISCNRecord[]
-      }
-      result = result
-        .concat(txRecords)
-        .concat(idRes ? idRes.records : [])
-        .concat(fingerprintRes ? fingerprintRes.records : [])
-        .concat(ownerRes ? ownerRes.records : [])
+      const { data } = await axios.get(`${network.apiURL}/iscn/records?q=${keyword}`)
```
nwingt, line 126: Seems can use axios that provides by the store without extra import
https://axios.nuxtjs.org/usage/#store-actions

lancatlin: I have tried `this.$axios`, `$axios`, `this.axios` but none of them are defined. 🤔 There isn't other usage of axios in `store` for reference 🥵

williamchong: would it be package version issue?

williamchong (approved):

PR #170, opened 2022-07-27
💄 Update purchase UI

AuroraHuang22:
截圖 2022-07-28 上午12 21 01

PR #176, opened 2022-07-28
🚸 Show owning count before collect

nwingt:
Based on #170 

Show owning count before collect
image


pages/nft/purchase/_iscnId.vue
```diff
@@ -130,15 +137,31 @@ export default class NFTTestButtonPage extends Vue {
...
       this.getPurchaseInfo(),
       this.getNftMetadata(),
+      this.getNFTHistory(),
     ]).catch((err) => this.$nuxt.error({ statusCode: 404, message: err }))
     await this.getOwnerName().catch(err => console.error(err))
     this.getLIKEPrice().catch(err => console.error(err))
   }
 
+  async getNFTHistory() {
+    const { data } = await this.$axios.get(API_LIKER_NFT_HISTORY, {
```
williamchong, line 156: Can we use chain API instead? less issue more robust (not sure if faster)
https://github.com/likecoin/liker-land/blob/feature/nft/src/pages/nft/class/_classId.vue#L449

williamchong (changes requested):

williamchong (approved):

PR #177, opened 2022-07-28
🚸[nft/purchase] Remove footer

AuroraHuang22:


layouts/wallet.vue
```diff
@@ -3,7 +3,9 @@
     <template v-if="currentAddress">
       <AppHeader />
       <Nuxt class="min-h-full" />
-      <AppFooter/>
+      <AppFooter
+        :class="{'hidden': $nuxt.$route.path.includes('/nft/purchase/')}"
```
nwingt, line 7:
```suggestion
        :class="{ hidden: $route.path.includes('/nft/purchase/')}"
```

williamchong: why not use name, or do we expect subpath under `/nft/purchase`? But in that case we can still use `name.includes`

nwingt (commented):
Isnt $route is accessible at this?

nwingt (approved):

PR #182, opened 2022-08-03
🥅 [nft/purchase] Guard insufficient balance

AuroraHuang22:
- Guard insufficient balance
- Display error response message from server side
截圖 2022-08-03 上午11 58 58


williamchong:
Missed this pr..
Can we use this pattern for display error instead? https://github.com/likecoin/app-like-co/pull/183

williamchong (approved):

PR #184, opened 2022-08-03
✨ Support querying by keyword

williamchong:
@lancatlin

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -116,7 +116,12 @@ import { logTrackerEvent } from '~/utils/logger'
 
 const iscnModule = namespace('iscn')
 
-@Component
+@Component({
+  fetch({ redirect, query }) {
+    const { q } = query;
+    if (q) redirect({ name: 'search-keyword___en', query });
+  },
+})
 export default class IndexPage extends Vue {
   @iscnModule.Action queryISCNByKeyword!: (
     arg0: string
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -117,9 +117,9 @@ import { logTrackerEvent } from '~/utils/logger'
 const iscnModule = namespace('iscn')
 
 @Component({
-  fetch({ redirect, query }) {
+  fetch({ redirect, query, localeLocation }) {
     const { q } = query;
-    if (q) redirect({ name: 'search-keyword___en', query });
+    if (q) redirect(localeLocation({ name: 'search-keyword', query })!);
   },
 })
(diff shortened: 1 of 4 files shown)


store/iscn.ts
```diff
@@ -161,6 +163,25 @@ export default class ISCN extends VuexModule {
...
 
+  @Action
+  async queryISCNByField({
+    keyword,
+  }: { keyword: string }): Promise<ISCNRecordWithID[]> {
+    this.context.commit('clearRecords');
+    let result: ISCNRecord[] = ([] as ISCNRecord[])
+    try {
+      this.context.commit('setIsLoading', true)
+      const { data } = await axios.get(`${network.apiURL}/iscn/records?keyword=${encodeURIComponent(keyword)}`)
```
lancatlin, line 174: I think it should be `/iscn/keywords?q=` for search all function (includes iscn id, owner, stakeholder, keywords). `keyword` will only search `contentMetadata.keywords`.

pages/index.vue
```diff
@@ -116,7 +116,12 @@ import { logTrackerEvent } from '~/utils/logger'
 
 const iscnModule = namespace('iscn')
 
-@Component
+@Component({
+  fetch({ redirect, query }) {
+    const { q } = query;
+    if (q) redirect({ name: 'search-keyword___en', query });
```
nwingt, line 122: can use `localeRoute({ name: 'search-keyword' })` ?

store/iscn.ts
williamchong: yes, the requirement was to specifically query by `keyword`. I think `queryISCNByKeyword`(poor name btw) above should be using `q` for search-all, and this function is for specific fields
https://github.com/likecoin/app-like-co/pull/175 should replace the above function by `q` after we fix performance issue (I based this PR on #175 actually)

pages/index.vue
williamchong: typescript doesn't have `localeLocation` in context, how can I hack it..

tsconfig.json
```diff
@@ -27,11 +27,11 @@
       ]
     },
     "types": [
+      "@types/node",
       "@nuxt/types",
+      "@nuxtjs/sentry",
```
nwingt, line 32: Why changing the order?

williamchong: it seems that the type below (axios, i18n) extends the type above (node => nuxt) so the order some how matters

nwingt (approved):

PR #187, opened 2022-08-04
🥅 Guard decay info and airdrop end message

WeiJun0827:


williamchong (approved):

nwingt (approved):

nwingt (approved):

locales/en.json
```diff
@@ -62,7 +63,7 @@
   "AirDrop.mission.button.previous": "Previous",
   "AirDrop.mission.button.retry": "retry",
   "AirDrop.mission.button.seeTransaction": "Check Tx",
-  "AirDrop.mission.button.technicalError": "Please try again in a few minutes.",
+  "AirDrop.mission.button.technicalError": "Not opened for claim",
```
nwingt, line 66:
```suggestion
  "AirDrop.mission.button.technicalError": "Not open to claim",
```

pages/airdrop/claim.vue
```diff
@@ -303,7 +315,7 @@ export default class AirdropClaimPage extends Vue {
     this.missionLoadingStatus = 'Loading'
     const res: any = await this.$axios
       .post(
```
williamchong, line 317: Can we block this action if airdrop end is true?

williamchong (changes requested):

pages/airdrop/claim.vue
WeiJun0827: Blocked, also blocked from the server

pages/airdrop/claim.vue
```diff
@@ -161,8 +159,12 @@ export default class AirdropClaimPage extends Vue {
     return `${BIG_DIPPER_TX_BASE_URL}${this.currentMission.txHash}`
   }
 
-  mounted() {
-    this.fetchMissionDecay()
+  get isAirdropEnd() {
+    return this.decay.factor === 1;
```
williamchong, line 163:
```suggestion
    return this.decay.factor >= 1;
```
In case the value somehow > 1

williamchong (changes requested):

williamchong (approved):

PR #190, opened 2022-08-05
🥅 Retry fetching `arweaveId` if submititon failed

WeiJun0827:


pages/nft/iscn/_iscnId.vue
```diff
@@ -292,7 +293,18 @@ export default class NFTTestMintPage extends Vue {
...
           if (!this.ogImageArweaveId) {
             const txHash = await this.sendArweaveFeeTx(arweaveFeeInfo)
-            await this.submitToArweave(txHash)
+            const TRY_LIMIT = 5
+            let tryTime = 0
+            do {
+              tryTime += 1
+              /* eslint-disable no-await-in-loop */
+              try {
+                await this.submitToArweave(txHash)
```
williamchong, line 302: In what case would `submitToArweave` fail, would we have submited multiple times (and spent multiple AR) if its only a timeout/delay?

pages/nft/iscn/_iscnId.vue
```diff
@@ -292,7 +293,18 @@ export default class NFTTestMintPage extends Vue {
           const arweaveFeeInfo = await this.estimateArweaveFee()
           if (!this.ogImageArweaveId) {
             const txHash = await this.sendArweaveFeeTx(arweaveFeeInfo)
-            await this.submitToArweave(txHash)
+            const TRY_LIMIT = 5
```
nwingt, line 296: Can it put outside of the class?

pages/nft/iscn/_iscnId.vue
WeiJun0827: I think the [API](https://github.com/likecoin/likecoin-api-public/blob/master/src/routes/arweave/index.js#L63) may fail for many reasons. Do we only aim at the `TX_NOT_FOUND` [case](https://github.com/likecoin/likecoin-api-public/blob/master/src/routes/arweave/index.js#L107)?

williamchong: Yes we want to keep querying if TX_NOT_FOUND, but we don't want to keep posting to AR since we might spend 3x the AR without knowing

WeiJun0827: It seems that I referred wrong code in my last reply, the frontend call app.like.co itself's API but not public API.
They seem smither with each other, I wonder why not just host the Arweave API on public API?

Also, I expected the response timeout status code should be `408`. And I tried to make the `POST /arweave/upload` sleep for 5 mins before it returns `arweaveId`, but I still cannot get a timeout error from that. So I didn't include the conditional check for a timeout in the end.

williamchong: well it was legacy issue, we had api in applikeco first, then moved to api public
I think can point applikeco to api public
what do you get for timeout now? 502/timeout?

WeiJun0827: Timeout didn't happen even I let `POST /arweave/upload` sleep for 5 mins before returning response.

williamchong: did it just wait 5min or sth else happened then?

williamchong: i think a smarter way to handle this loop is to just handle this in api, since we can specifically loop for arweave tx in api
in client, a better ux is to store the (successful) tx hash, and let use click to trigger retry in case sth happen

williamchong (changes requested):

pages/nft/iscn/_iscnId.vue
WeiJun0827: It just waited for 5 mins 🌚

pages/nft/url/index.vue
```diff
@@ -240,13 +239,15 @@ export default class FetchIndex extends Vue {
...
+      this.arweaveFeeTxHash = ''
       await this.crawlUrlData()
       if (this.crawledData?.body) {
         try {
           const arweaveFeeInfo = await this.estimateArweaveFee()
           if (!this.arweaveId) {
-            const txHash = await this.sendArweaveFeeTx(arweaveFeeInfo)
-            await this.submitToArweave(txHash)
+            await this.sendArweaveFeeTx(arweaveFeeInfo)
+            await this.submitToArweave()
```
williamchong, line 250: Should we skip `sendArweaveFeeTx` if `arweaveFeeTxHash ` is set?
Would `arweaveFeeTxHash ` reset if we get api error that thinks its invalid?

Similarly, would we skip `crawlUrlData` if `this.crawledData?.body` exists?

williamchong (changes requested):

williamchong (approved):
Probably need some QA testing on the UI recover/reset cases on testnet before deploy

PR #191, opened 2022-08-06
✨ Add user ts in purchase

Tsai-Hsueh-Kuan:
✨ Add user ts in purchase


pages/nft/purchase/_iscnId.vue
```diff
@@ -272,6 +272,7 @@ export default class NFTTestButtonPage extends Vue {
           iscn_id: this.iscnId,
           class_id: this.classId,
           tx_hash: this.grantTransactionHash,
```
williamchong, line 274: i think just `ts` would do

williamchong (changes requested):

williamchong (approved):

PR #192, opened 2022-08-08
✨ Add origin img src in crawler

Tsai-Hsueh-Kuan:
✨ Add origin img src in crawler

### DIFF
--- a/server/crawler/index.ts
+++ b/server/crawler/index.ts
@@ -210,7 +210,7 @@ export default async function getCralwerData(url: string) {
     const img = $('img')
     Object.keys(img).forEach((key: any) => {
       const { src } = img[key].attribs || {};
-      if(!src){
+      if (!src) {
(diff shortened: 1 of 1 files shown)


server/crawler/index.ts
```diff
@@ -201,14 +201,20 @@ export default async function getCralwerData(url: string) {
...
       'svg',
     ].forEach((t: string) => $(t).remove())
     title = $('title').text()
     const metas = $('meta')
     body = $('body').html() || ''
 
+    const img = $('img')
+    Object.keys(img).forEach((key: any) => {
+      const { src } = img[key].attribs || {};
+      if(!src){
```
williamchong, line 213:
```suggestion
      if (!src) {
```

williamchong (changes requested):

williamchong (approved):

PR #200, opened 2022-08-15
Feat/upload file img

Tsai-Hsueh-Kuan:
Feat/upload file img

### DIFF
--- a/pages/nft/url/index.vue
+++ b/pages/nft/url/index.vue
@@ -89,7 +89,7 @@ import {
   API_POST_ARWEAVE_UPLOAD,
 } from '~/constant/api'
 
-const base64toBlob = (base64Data:string, contentType = 'image/jpeg', sliceSize = 512) => {
+const base64toBlob = (base64Data:string, contentType: string, sliceSize = 512) => {
   const byteCharacters = atob(base64Data);
   const byteArrays = [];
   for (let offset = 0; offset < byteCharacters.length; offset += sliceSize) {
@@ -143,9 +143,9 @@ export default class FetchIndex extends Vue {
     const body = new Blob([this.crawledData.body], { type: "text/html" })
     const formData = new FormData()
     formData.append('index.html', body, 'index.html')
-    const { imgDataKey } = this.crawledData
-    imgDataKey.forEach((element:any) => {
-      formData.append(`${element.key}`, base64toBlob(element.data), `${element.key}`)
+    const { images } = this.crawledData
+    images.forEach((element:any) => {
+      formData.append(`${element.key}`, base64toBlob(element.data, element.type), `${element.key}`)
     });
     return formData
   }
--- a/pages/nft/url/index.vue
+++ b/pages/nft/url/index.vue
@@ -353,6 +353,7 @@ export default class FetchIndex extends Vue {
           headers: {
             'Content-Type': 'multipart/form-data',
           },
+          timeout: 180000,
         },
       )
       this.arweaveId = arweaveId
--- a/pages/nft/url/index.vue
+++ b/pages/nft/url/index.vue
@@ -130,7 +130,7 @@ export default class FetchIndex extends Vue {
     formData.append('file', body, 'index.html')
     const {imgDataKey} = this.crawledData
     imgDataKey.forEach((element:any) => {
-      formData.append(`${element.key}`, new Blob([element.data], {type: "image/png"}), `${element.key}`)
+      formData.append(`${element.key}`, new File([element.data],`${element.key}`), `${element.key}`)
     });
     return formData
   }
--- a/pages/nft/url/index.vue
+++ b/pages/nft/url/index.vue
@@ -179,8 +179,8 @@ export default class FetchIndex extends Vue {
 
   get iscnParams() {
     const params: any = { iscnId: this.iscnId }
-    if (this.crawledData?.image) {
-      params.ogImageUrl = this.crawledData.image
+    if (this.crawledData?.ogImage) {
+      params.ogImageUrl = this.crawledData.ogImage
     }
     return params
   }
--- a/pages/nft/url/index.vue
+++ b/pages/nft/url/index.vue
@@ -128,8 +128,8 @@ export default class FetchIndex extends Vue {
     const body = new Blob([this.crawledData.body], { type: "text/html" })
     const formData = new FormData()
     formData.append('file', body, 'index.html')
-    const a = this.crawledData.imgBase64
-    a.forEach((element:any) => {
+    const {imgDataKey} = this.crawledData
+    imgDataKey.forEach((element:any) => {
       formData.append(`${element.key}`, new Blob([element.data], {type: "image/png"}), `${element.key}`)
     });
     return formData
--- a/pages/nft/url/index.vue
+++ b/pages/nft/url/index.vue
@@ -353,7 +353,7 @@ export default class FetchIndex extends Vue {
           headers: {
             'Content-Type': 'multipart/form-data',
           },
-          timeout: 180000,
+          timeout: 90000,
         },
       )
       this.arweaveId = arweaveId
--- a/pages/nft/url/index.vue
+++ b/pages/nft/url/index.vue
@@ -128,6 +128,10 @@ export default class FetchIndex extends Vue {
     const body = new Blob([this.crawledData.body], { type: "text/html" })
     const formData = new FormData()
     formData.append('file', body, 'index.html')
+    const a = this.crawledData.imgBase64
+    a.forEach((element:any) => {
+      formData.append(`${element.key}`, new Blob([element.data], {type: "image/png"}), `${element.key}`)
+    });
     return formData
   }
 
--- a/pages/nft/url/index.vue
+++ b/pages/nft/url/index.vue
@@ -89,6 +89,25 @@ import {
   API_POST_ARWEAVE_UPLOAD,
 } from '~/constant/api'
 
+const b64toBlob = (b64Data:string, contentType='image/jpeg', sliceSize=512) => {
+  const byteCharacters = atob(b64Data);
+  const byteArrays = [];
+
+  for (let offset = 0; offset < byteCharacters.length; offset += sliceSize) {
+    const slice = byteCharacters.slice(offset, offset + sliceSize);
+
+    const byteNumbers = new Array(slice.length);
+    for (let i = 0; i < slice.length; i+=1) {
+      byteNumbers[i] = slice.charCodeAt(i);
+    }
+
+    const byteArray = new Uint8Array(byteNumbers);
+    byteArrays.push(byteArray);
+  }
+
+  const blob = new Blob(byteArrays, {type: contentType});
+  return blob;
+}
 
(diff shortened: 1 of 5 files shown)


server/arweave/index.ts
```diff
@@ -177,10 +175,12 @@ export async function convertARPriceToLIKE(ar: ArweavePrice, {
 export async function convertARPricesToLIKE(ar: ArweavePrice,
   { margin = 0.05, decimal = 0 } = {},
 ): Promise<ArweavePriceWithLIKE> {
+  const { data } = await axios.get(COINGECKO_PRICE_API);
+  const { likecoin, arweave: arweavePrice } = data;
   if (!(ar.list && ar.list.length)) {
-    return convertARPriceToLIKE(ar, { margin, decimal });
+    return convertARPriceToLIKE(ar, { margin, decimal }, likecoin, arweavePrice);
```
williamchong, line 181: instead of adding `likecoin`, can we calculate `priceRatio` as input for `convertARPriceToLIKE` and keep the original function interface instead?

pages/nft/url/index.vue
```diff
@@ -127,7 +142,11 @@ export default class FetchIndex extends Vue {
     if (!this.crawledData?.body) { return null }
     const body = new Blob([this.crawledData.body], { type: "text/html" })
     const formData = new FormData()
-    formData.append('file', body, 'index.html')
+    formData.append('index.html', body, 'index.html')
+    const { imgDataKey } = this.crawledData
+    imgDataKey.forEach((element:any) => {
+      formData.append(`${element.key}`, base64toBlob(element.data), `${element.key}`)
```
williamchong, line 148: Can we add the `contentType` from original blob instead of hardcode `jpeg`?

server/crawler/index.ts
```diff
@@ -195,25 +195,79 @@ export default async function getCralwerData(url: string) {
...
+    //     return element
+    //   })
+    //   .catch(()=> {}));
+    // }})
+    // const sourceData:any = await Promise.all(promiseSource)
+    // for (let j = 0; j < sourceData.length; j+=1 ) {
+    //   if(sourceData[j]?.status === 200){
+    //     imgDataKey.push( {'data': Buffer.from(sourceData[j].data, 'binary').toString('base64'), 'key': sourceKey[j]} );
+    //   }
+    // }
```
williamchong, line 270: Remove if unused

server/crawler/index.ts
```diff
@@ -232,5 +286,5 @@ export default async function getCralwerData(url: string) {
     // eslint-disable-next-line no-console
     console.error(error)
   }
-  return { title, description, keywords, author, body, image }
+  return { title, description, keywords, author, body, image, imgDataKey }
```
williamchong, line 289: Rename `image` to `ogImage`, `imgDataKey` to `images`?

server/crawler/index.ts
```diff
@@ -195,25 +195,79 @@ export default async function getCralwerData(url: string) {
...
+          imgKey.push(`./img${key}.png`)
+          return element
+        })
+        .catch(()=> {}));
       }
     })
+    const imgData:any = await Promise.all(promiseImg)
+    for (let i = 0; i < imgData.length; i+=1 ){
+      if(imgData[i]?.status === 200){
+        imgDataKey.push({ 'data': Buffer.from(imgData[i].data, 'binary').toString('base64'), 'key': imgKey[i]} );
```
williamchong, line 242: I think we want to store mimetype too

server/crawler/index.ts
```diff
@@ -195,25 +195,79 @@ export default async function getCralwerData(url: string) {
...
+        let srcUrl = src
+        if ( !src.startsWith('http') ) {
+          srcUrl = `${protocol}//${host}${src}`
+        }
+        promiseImg.push(axios.get(`${srcUrl}`, {responseType: 'arraybuffer'})
+        .then((element)=> {
+          const srcChange = src.replace(/&/g, `&amp;`)
+          body = body.replace(src, `./img${key}.png`)
+          body = body.replace(srcChange, `./img${key}.png`)
+          imgKey.push(`./img${key}.png`)
```
williamchong, line 233: Can we get the original `mime/type` instead of hard code `png`?

server/crawler/index.ts
```diff
@@ -195,25 +195,79 @@ export default async function getCralwerData(url: string) {
...
       if (!src) {
         delete img[key]
+      } else {
+        let srcUrl = src
+        if ( !src.startsWith('http') ) {
+          srcUrl = `${protocol}//${host}${src}`
+        }
+        promiseImg.push(axios.get(`${srcUrl}`, {responseType: 'arraybuffer'})
+        .then((element)=> {
+          const srcChange = src.replace(/&/g, `&amp;`)
```
williamchong, line 230: Would using `.map` + `Promise.all` get us a simpler structure than `.forEach` and `.then`?

server/crawler/index.ts
```diff
@@ -195,25 +195,79 @@ export default async function getCralwerData(url: string) {
...
+          body = body.replace(srcChange, `./img${key}.png`)
+          imgKey.push(`./img${key}.png`)
+          return element
+        })
+        .catch(()=> {}));
       }
     })
+    const imgData:any = await Promise.all(promiseImg)
+    for (let i = 0; i < imgData.length; i+=1 ){
+      if(imgData[i]?.status === 200){
```
williamchong, line 241: just use `.map` + `.filter` ?

williamchong (changes requested):

server/arweave/index.ts
```diff
@@ -150,15 +150,17 @@ export async function estimateARPrices(files: ArweaveFile[]): Promise<ArweavePri
...
+async function getPriceRatioBigNumber () {
+  const { data } = await axios.get(COINGECKO_PRICE_API);
+  const { likecoin, arweave: arweavePrice } = data;
+  const priceRatio = new BigNumber(arweavePrice.usd).dividedBy(likecoin.usd).toFixed();
   const priceRatioBigNumber = BigNumber.max(priceRatio, 1);
+  return priceRatioBigNumber;
+}
+
+export function convertARPriceToLIKE(ar: ArweavePrice, {
+  priceRatioBigNumber = 1 as any, margin = 0.05, decimal = 0,
```
williamchong, line 162: `priceRatioBigNumber` should not have default value, calling this with value `1` would cause serious calculation mistake. It should call `getPriceRatioBigNumber` if one is not set

server/crawler/index.ts
```diff
@@ -228,9 +226,15 @@ export default async function getCralwerData(url: string) {
...
-          imgKey.push(`./img${key}.png`)
+          let extension
+          if (src.endsWith('jpeg')) {
+            extension = 'jpeg'
+          } else if(src.endsWith('png')) {
+            extension = 'png'
+          }
+          body = body.replace(src, `./img${key}.${extension}`)
+          body = body.replace(srcChange, `./img${key}.${extension}`)
+          imgKey.push(`./img${key}.${extension}`)
```
williamchong: Can we just keep the original filename/path without adding `.jpeg` or .`png`?

williamchong (changes requested):

server/crawler/index.ts
```diff
@@ -194,25 +194,52 @@ export default async function getCralwerData(url: string) {
...
+          $(this).attr('src', newFileName); 
+          return { element, key: newFileName }
+        })
+        .catch(()=> {}));
       }
+    });
+
+    const imgData:any = await Promise.all(promiseImg)
+    imgData.filter((e: any) => e?.element?.status === 200).map((e: any) => {
+      images.push({ 'data': Buffer.from(e.element.data, 'binary').toString('base64'), 'key': e.key, 'type': e.element.headers['content-type']} );
```
nnkken, line 241: Separate the object into multiple lines for readability, also the quotes around keys are not needed

server/crawler/index.ts
```diff
@@ -194,26 +194,58 @@ export default async function getCralwerData(url: string) {
...
+    const promiseImg:any = []
 
+    let index = 0
+    const urlIndexObject:any = {}
     const img = $('img')
-    Object.keys(img).forEach((key: any) => {
-      const { src } = img[key].attribs || {};
-      if (!src) {
-        delete img[key]
+    img.each(function() {
```
williamchong, line 217: 應該each 本身就內建了 i 可以直接用？
https://cheerio.js.org/classes/Cheerio.html#each

williamchong (changes requested):

server/crawler/index.ts
```diff
@@ -194,26 +194,58 @@ export default async function getCralwerData(url: string) {
...
+          srcUrl = `${protocol}//${host}${src}`
+        }
+        if(!urlIndexObject[srcUrl]){
+          urlIndexObject[srcUrl] = index
+          index += 1
+        }
+        promiseImg.push(axios.get(`${srcUrl}`, {responseType: 'arraybuffer'})
+        .then((element)=> {
+          const elementUrl = element?.config?.url
+          const newFileName = `./img${urlIndexObject[`${elementUrl}`]}`
```
williamchong, line 231: Also can we name them as `/images/{i}` for clarity?

williamchong (changes requested):

server/crawler/index.ts
Tsai-Hsueh-Kuan: If there is `/` in the middle, there will be problems.
i must use the object to save it first, otherwise, when the promise is doing, it dosnt know what it is?

williamchong: what kind of problem exactly?

server/crawler/index.ts
```diff
@@ -194,26 +194,51 @@ export default async function getCralwerData(url: string) {
...
+    const imgData:any = await Promise.all(promiseImg)
+    imgData.filter( (e: any) => e?.element?.status === 200 )
+      .map((e: any) => {
+        images.push({ 
+          data: Buffer.from(e.element.data, 'binary').toString('base64'), 
+          key: e.key, 
+          type: e.element.headers['content-type'],
+        });
+        return ''
+      })
```
williamchong, line 241: Should not use .push when we are using .map
```
const images = imgData.filter( (e: any) => e?.element?.status === 200 )
      .map(...)
```
.map returns an array, which should be content of images.

pages/nft/url/index.vue
```diff
@@ -89,6 +89,21 @@ import {
   API_POST_ARWEAVE_UPLOAD,
 } from '~/constant/api'
 
+const base64toBlob = (base64Data:string, contentType: string, sliceSize = 512) => {
```
williamchong, line 92: What does `sliceSize` mean here? Would 512 always work?

williamchong (changes requested):

pages/nft/url/index.vue
Tsai-Hsueh-Kuan: I mainly refer to the Internet and see that many people use 512
https://stackoverflow.com/questions/16245767/creating-a-blob-from-a-base64-string-in-javascript
https://www.796t.com/post/MTF4YQ==.html

williamchong (approved):

PR #202, opened 2022-08-16
💬 Update alert message for keplr not found

AuroraHuang22:
截圖 2022-08-16 上午11 54 40


ckxpress:
for "detailed" installation guide

p.s. so happen to pass by

------- Original Message -------

nwingt (approved):

PR #203, opened 2022-08-17
✨ Support retrying & skipping features for ISCN registration and format error handling structure

WeiJun0827:
- `console.error(error)` and throw user-friendly `ERROR_MESSAGE` when catching an error in child functions.
- `setError(ERROR_MESSAGE)` when catching an error in functions that bonded with UI components. 
- Control ISCN registration flow with state pattern.

### DIFF
--- a/pages/nft/iscn/_iscnId.vue
+++ b/pages/nft/iscn/_iscnId.vue
@@ -352,7 +352,7 @@ export default class NFTTestMintPage extends Vue {
         if (!image) return
         this.ogImageUrl = image
       }
-      const { data, headers } = await this.$axios.get(this.ogImageUrl)
+      const { data, headers } = await this.$axios.get(`/crawler/image?url=${encodeURIComponent(this.ogImageUrl)}`)
       this.ogImageBlob = new Blob([data], { type: headers['content-type'] })
     } catch (error) {
       // eslint-disable-next-line no-console
--- a/pages/nft/iscn/_iscnId.vue
+++ b/pages/nft/iscn/_iscnId.vue
@@ -246,7 +246,7 @@ export default class NFTTestMintPage extends Vue {
         this.getISCNInfo(),
         this.getMintInfo(),
       ])
-      this.getOgImage()
+      await this.getOgImage()
     } catch (error) {
       this.setError(error)
     }
--- a/pages/nft/iscn/_iscnId.vue
+++ b/pages/nft/iscn/_iscnId.vue
@@ -241,70 +241,82 @@ export default class NFTTestMintPage extends Vue {
   }
 
   async mounted() {
-    await Promise.all([
-      this.getISCNInfo().catch(err => {
-        // eslint-disable-next-line no-console
-        console.error(err);
-        this.setError('ISCN_NOT_FOUND')
-      }),
-      // eslint-disable-next-line no-console
-      this.getMintInfo().catch(err => console.error(err)),
-    ]);
-    this.getOgImage()
+    try {
+      await Promise.all([
+        this.getISCNInfo(),
+        this.getMintInfo(),
+      ])
+      this.getOgImage()
+    } catch (error) {
+      this.setError(error)
+    }
   }
 
   async doAction() {
-    this.errorType = ''
-    this.balance = await getAccountBalance(this.address) as string
-    if(this.balance === '0'){
-      this.toggleSnackbar(ErrorType.INSUFFICIENT_BALANCE)
-      return
-    }
-    this.isLoading = true
-    /* eslint-disable no-fallthrough */
-    switch (this.state) {
-      case 'create':
-        if (!this.isUserISCNOwner) {
-          this.toggleSnackbar(ErrorType.USER_NOT_ISCN_OWNER)
-          break
-        }
+    try {
+      this.errorType = ''
+      this.balance = await getAccountBalance(this.address) as string
+      if (this.balance === '0'){
+        throw new Error(ErrorType.INSUFFICIENT_BALANCE)
+      }
+      this.isLoading = true
+      /* eslint-disable no-fallthrough */
+      switch (this.state) {
+        case 'create':
+          if (!this.isUserISCNOwner) {
+            throw new Error(ErrorType.USER_NOT_ISCN_OWNER)
+          }
 
-        try {
           if (this.ogImageBlob) {
-            const arweaveFeeInfo = await this.estimateArweaveFee()
-            if (!this.ogImageArweaveId) {
-              if (!this.ogImageArweaveFeeTxHash) { await this.sendArweaveFeeTx(arweaveFeeInfo) }
-              await this.submitToArweave()
+            try {
+              const arweaveFeeInfo = await this.estimateArweaveFee()
+              if (!this.ogImageArweaveId) {
+                if (!this.ogImageArweaveFeeTxHash) { await this.sendArweaveFeeTx(arweaveFeeInfo) }
+                await this.submitToArweave()
+              }
+            } catch (error) {
+              // eslint-disable-next-line no-console
+              console.error(error);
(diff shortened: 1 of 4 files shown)


pages/nft/iscn/_iscnId.vue
```diff
@@ -241,70 +241,82 @@ export default class NFTTestMintPage extends Vue {
...
+            try {
+              const arweaveFeeInfo = await this.estimateArweaveFee()
+              if (!this.ogImageArweaveId) {
+                if (!this.ogImageArweaveFeeTxHash) { await this.sendArweaveFeeTx(arweaveFeeInfo) }
+                await this.submitToArweave()
+              }
+            } catch (error) {
+              // eslint-disable-next-line no-console
+              console.error(error);
+              // skip uploading image if there is an error
```
williamchong, line 280: I think should keep showing the error if `ogImageBlob` exists, unless we have a UI to allow user to skip uploading

pages/nft/iscn/_iscnId.vue
```diff
@@ -320,15 +332,10 @@ export default class NFTTestMintPage extends Vue {
...
-            data,
-          }), this.redirectOrigin);
-        } catch (error) {
-          // eslint-disable-next-line no-console
-          console.error(error);
-        }
+        window.opener.postMessage(JSON.stringify({
+          action: 'NFT_MINT_DATA',
+          data,
+        }), this.redirectOrigin)
```
williamchong, line 338: Why remove the try catch??

pages/nft/iscn/_iscnId.vue
```diff
@@ -241,70 +241,82 @@ export default class NFTTestMintPage extends Vue {
...
-        break
-      case 'done':
-        window.location.href = this.detailsPageURL
-      default:
+  
+        case 'mint':
+          await this.mintNFT()
+          if (!this.sendRes) break
+          if (this.isWritingNFT) {
+            await this.postMintInfo()
```
williamchong, line 290: I think the original way to write is clearer, you see `this.sendRes` assignment and its use in single screen
Any reason why you want to put it inside?

pages/nft/url/index.vue
```diff
@@ -207,57 +219,82 @@ export default class FetchIndex extends Vue {
...
+        if (!this.url) {
+          this.errorMessage = this.$t('HomePage.search.errormessage.empty') as string
+          break
+        }
+        this.state = State.TO_CRAWL
+        break
+      case State.TO_CRAWL:
+        await this.crawlUrlData()
+        this.state = State.TO_ESTIMATE_FEE
+        break
```
williamchong, line 278: If we don't `break` after each sucessful `State` and only break on error, we don't need to use a while loop, just call this function will do

pages/nft/url/index.vue
```diff
@@ -290,28 +328,29 @@ export default class FetchIndex extends Vue {
...
-      console.error('CANNOT_ESTIMATE_ARWEAVE_FEE')
-      throw err
+      console.error(err)
+      throw new Error('CANNOT_ESTIMATE_ARWEAVE_FEE')
     }
   }
 
-  async sendArweaveFeeTx({ to, amount, memo }: { to: string, amount: BigNumber, memo: string }): Promise<void> {
+  async sendArweaveFeeTx(): Promise<void> {
+    const { to, amount, memo } = this.arweaveFeeInfo
```
williamchong, line 344: Personally I think it's more readable if you read and write `arweaveFeeInfo` in the `doAction` loop, instead of the new change here

williamchong (changes requested):

pages/nft/iscn/_iscnId.vue
WeiJun0827: Will the try catch outside of this one do the same job?

pages/nft/iscn/_iscnId.vue
WeiJun0827: Cuz I see other functions tend to assign props(`this.xxx`) inside the function., and inevitably check these props outside in the caller functions.
So I try to follow the style when refactoring.

pages/nft/url/index.vue
WeiJun0827: Yes, I agree. But currently, the read/write of the original `arweaveFeeInfo` happens in different case blocks, so I think I have to store it in prop temporarily.

pages/nft/url/index.vue
WeiJun0827: Yes, that's smarter!

pages/nft/iscn/_iscnId.vue
WeiJun0827: I think an easier way is to pop up a snackbar to warn user, but a more complete way is to implement a switch fall-through pattern similar to the ISCN registration flow to let users be able to retry?

pages/nft/iscn/_iscnId.vue
williamchong: not really, I explicitly want to catch any error related to `window.opener.postMessage`, which `window.opener` is something we cannot control and might yield ignore-able error on postMessage

pages/nft/url/index.vue
williamchong: Where do you use `sendArweaveFeeTx` except in `doAction`? If we only use it in do action, I think the `this.arweaveFeeInfo` can happen in doAction instead of here

pages/nft/iscn/_iscnId.vue
williamchong: I think popup a snackbar is fine
The issue would be we need to either allow user to retry the call, or explicitly do sth to skip this step, instead of default skipping on error

pages/nft/iscn/_iscnId.vue
williamchong: at lease the prop needs to be named to be related to the method I guess? `sendRes` is too general and cause confusion. `mintNFTResult` would be much more understandable without context

server/crawler/router.ts
```diff
@@ -9,10 +9,24 @@ router.get('/', async (req, res, next) => {
...
+})
+
+router.get('/image', async (req, res, next) => {
+  try {
+    const { url } = req.query
+    if (!url) {
+      res.status(400).send('MISSING_IMAGE_URL')
+      return
+    }
+    const data = await crawlImage(url as string)
```
williamchong, line 28: Now this api allow getting any image from any url
I think it would be better to limit the api by using same url used in `/`, fetch if there is an ogImage, then return the ogImage if exists.
Also we should use stream and pipe for image data api

server/crawler/index.ts
```diff
@@ -257,3 +257,20 @@ export default async function getCralwerData(url: string) {
...
+  const metas = $('meta')
+  let ogImageUrl = ''
+  for (let i = 0; i < Object.keys(metas).length; i += 1) {
+    const { attribs } = metas[i]; // keys are integers 1 to N
+    if (attribs?.property === 'og:image') {
+      ogImageUrl = attribs.content
+      break
+    }
+  }
+  if (!ogImageUrl) return null;
```
williamchong, line 273: Just reuse `crawlData()` here?

williamchong (changes requested):

server/crawler/index.ts
WeiJun0827: Sure we can get all the necessary info in `crawlData()`, including ogImage. But because we don't need to cache ogImage URL when registering ISCN. I think we can actually remove searching ogImage logic in `crawlData()`. Also, `crawlData()` calls additional async functions like `Promise.all(promiseImg)` that creating class ID doesn't need. Is it preferred to separate the crawling logic into two functions?

williamchong: Hmm yes the function got complex after arweave image support... lets just use this version here then

williamchong (approved):

PR #212, opened 2022-08-18
✨ Make tag clickable to keywords search

williamchong:


AuroraHuang22 (approved):

nwingt (commented):
There might be non-keyword use cases for `<Tag>`.

I suggest modifying `<Tag>` using [`<component>`](https://v2.vuejs.org/v2/guide/components.html#Dynamic-Components) with default as `div` and creating a new component e.g. `<Keyword>` wrapping `<Tag>`.

nwingt (approved):

PR #213, opened 2022-08-19
📈 Log events for NFT related actions

AuroraHuang22:


williamchong (changes requested):
Please add event when user start `grantPurchaseTransaction` and after `signingClient.createSendGrant` successful

williamchong (changes requested):
Please also log before transactions success
ie when user trigger onClickMint but before signer send transaction succeed

PR #222, opened 2022-09-02
📈 Add mint NFT logTracker

Tsai-Hsueh-Kuan:
📈 Add mint NFT logTracker

### DIFF
--- a/pages/nft/iscn/_iscnId.vue
+++ b/pages/nft/iscn/_iscnId.vue
@@ -259,6 +259,7 @@ export default class NFTTestMintPage extends Vue {
       this.hasError = false
       this.balance = await getAccountBalance(this.address) as string
       if (this.balance === '0') {
+        logTrackerEvent(this, 'IscnMintNFT', 'GetAccountBalanceError', ErrorType.INSUFFICIENT_BALANCE, 1);
         throw new Error(ErrorType.INSUFFICIENT_BALANCE)
       }
       this.isLoading = true
@@ -267,11 +268,13 @@ export default class NFTTestMintPage extends Vue {
         case 'create': {
           const isAllowed = IS_TESTNET || await this.checkIsWhitelisted();
           if (!isAllowed) {
+            logTrackerEvent(this, 'IscnMintNFT', 'CheckIsWhitelistedError', ErrorType.USER_NOT_WHITELISTED, 1);
             this.toggleSnackbar(ErrorType.USER_NOT_WHITELISTED)
             break
           }
 
           if (!this.isUserISCNOwner) {
+            logTrackerEvent(this, 'IscnMintNFT', 'CheckIsWhitelistedError', ErrorType.USER_NOT_ISCN_OWNER, 1);
             throw new Error(ErrorType.USER_NOT_ISCN_OWNER)
           }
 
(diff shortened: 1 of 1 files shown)


pages/nft/iscn/_iscnId.vue
```diff
@@ -307,12 +308,14 @@ export default class NFTTestMintPage extends Vue {
   }
 
   async checkIsWhitelisted() {
+    logTrackerEvent(this, 'IscnMintNFT', 'CheckIsWhitelisted', this.address, 1);
```
williamchong, line 311: I think we should log on the error which is not whitelisted, logging `checkIsWhitelisted` will be spammy since every on submit trigger this now

pages/nft/iscn/_iscnId.vue
```diff
@@ -307,12 +308,14 @@ export default class NFTTestMintPage extends Vue {
...
     const { data } = await this.$axios.get(getWhitelistApi(this.address))
     return data.isWhitelisted;
   }
 
   async getISCNInfo() {
     try {
+      logTrackerEvent(this, 'IscnMintNFT', 'GetISCNInfo', this.iscnId, 1);
```
williamchong, line 318: I think no need to log this since this is auto triggered by `mounted`
We should log the `Warning: No URL in ISCN\'s metadata` below though

pages/nft/iscn/_iscnId.vue
```diff
@@ -330,6 +334,7 @@ export default class NFTTestMintPage extends Vue {
 
   async getMintInfo() {
     try {
+      logTrackerEvent(this, 'IscnMintNFT', 'GetMintInfo', this.iscnId, 1);
```
williamchong, line 337: Similar to above, I think no need to log the function trigger

williamchong (changes requested):
We should log error case like 
`ErrorType.INSUFFICIENT_BALANCE`
`ErrorType.USER_NOT_WHITELISTED`
`ErrorType.USER_NOT_ISCN_OWNER`
`!this.signer`
and `post/action` functions in `doAction()`
but we don't need to log get functions that are not triggered by new user action

pages/nft/iscn/_iscnId.vue
```diff
@@ -266,11 +268,13 @@ export default class NFTTestMintPage extends Vue {
         case 'create': {
           const isAllowed = IS_TESTNET || await this.checkIsWhitelisted();
           if (!isAllowed) {
+            logTrackerEvent(this, 'IscnMintNFT', 'CheckIsWhitelistedError', ErrorType.USER_NOT_WHITELISTED, 1);
             this.toggleSnackbar(ErrorType.USER_NOT_WHITELISTED)
             break
           }
 
           if (!this.isUserISCNOwner) {
+            logTrackerEvent(this, 'IscnMintNFT', 'CheckIsWhitelistedError', ErrorType.USER_NOT_ISCN_OWNER, 1);
```
williamchong, line 277: I think a better log format would be set both error as `CreateNFTError` as `action` then `ErrorType.xx` as value
`GetAccountBalanceError` in above should be changed similarly

pages/nft/iscn/_iscnId.vue
```diff
@@ -359,10 +367,12 @@ export default class NFTTestMintPage extends Vue {
   async getOgImage() {
     try {
       const url = this.iscnData.contentMetadata?.url
+      logTrackerEvent(this, 'IscnMintNFT', 'GetOgImage', url, 1);
       if (!url) return
```
williamchong, line 371: I think we would want to split log for url exists and not exists case

pages/nft/iscn/_iscnId.vue
```diff
@@ -387,19 +398,28 @@ export default class NFTTestMintPage extends Vue {
...
-    if (!this.signer) throw new Error('SIGNER_NOT_INITED')
-    if (!to) throw new Error('TARGET_ADDRESS_NOT_SET')
+    if (!this.signer) {
+      logTrackerEvent(this, 'IscnMintNFT', 'SendArweaveFeeTxSignerError', 'SIGNER_NOT_INITED', 1);
+      throw new Error('SIGNER_NOT_INITED')
+    }
+    if (!to) {
+      logTrackerEvent(this, 'IscnMintNFT', 'SendArweaveFeeTxTargetError', 'TARGET_ADDRESS_NOT_SET', 1);
+      throw new Error('TARGET_ADDRESS_NOT_SET')
+    }
```
williamchong, line 416: We can set all log action as `SendArweaveFeeTxError` and use different label for each cases

williamchong (changes requested):

williamchong (approved):

PR #224, opened 2022-09-02
✨ Integrate keyword search function in detail page

WeiJun0827:


pages/view/_iscnId.vue
```diff
@@ -332,14 +339,21 @@
                 :label="$t('iscn.meta.creator.name')"
                 class="w-[50%] my-[12px]"
               >
-                <Label :text="stakeholderInfo.authorName" tag="div" preset="p4" />
+                <Link
+                  v-if="stakeholderInfo.authorWalletAddresses.length"
+                  :class="'text-[16px]'"
+                  :href="`/search?q=${stakeholderInfo.authorWalletAddresses[0].address}`"
```
williamchong, line 345: can we use name route instead of URL, and use explicit query for each field instead of query All (q)?

pages/view/_iscnId.vue
```diff
@@ -178,7 +178,14 @@
             :label="$t('iscn.meta.owner')"
             class="mb-[12px]"
           >
-            <Label :text="owner" tag="div" preset="p6" />
+            <Link
+              :class="[
+                'text-[14px]',
+                'break-all',
+              ]"
+              :href="`/search?q=${owner}`">
```
williamchong, line 186: can we use name route instead of URL, and use explicit query for each field instead of query All (q)?

williamchong (changes requested):

pages/view/_iscnId.vue
williamchong: ie https://github.com/likecoin/app-like-co/blob/main/pages/search/_keyword.vue#L164 this function `queryISCNByField` only support `keyword` now but it support owner/stakholder query on API level, please help implement the support of these field and use this function instead of `q`

pages/search/_keyword.vue
```diff
@@ -153,14 +160,35 @@ export default class SearchPage extends Vue {
...
+        keyword,
+      } = this.$route.query;
       const searchObject = {
-        keyword: (keyword || keywords) as string,
+        iscnId: iscnId as string,
+        owner: owner as string,
+        contentFingerprint: contentFingerprint as string,
+        stakeholderId: stakeholderId as string,
+        stakeholderName: stakeholderName as string,
+        keyword: keyword as string,
```
WeiJun0827, line 191: To simplify the search object, I keep the `keyword` and delete the `keywords`.
For `stakeholderId`, `stakeholderName`, and `keyword`, it's possible to declare as `string[]` to set multiple conditions.
But since the use case is clicks and search, I use the single term for now.
Is it preferred to implement with plural?

pages/view/_iscnId.vue
```diff
@@ -332,14 +340,21 @@
                 :label="$t('iscn.meta.creator.name')"
                 class="w-[50%] my-[12px]"
               >
-                <Label :text="stakeholderInfo.authorName" tag="div" preset="p4" />
+                <Link
+                  v-if="stakeholderInfo.authorWalletAddresses.length"
```
williamchong, line 344: Why check for `stakeholderInfo.authorWalletAddresses.length` when the Link is about `stakeholderInfo.authorName`?

store/iscn.ts
```diff
@@ -136,13 +136,34 @@ export default class ISCN extends VuexModule {
...
       this.context.commit('setIsLoading', true)
-      const { data } = await axios.get(`${network.apiURL}/iscn/records?keyword=${encodeURIComponent(keyword)}`)
+
+      const url = new URL(`${network.apiURL}/iscn/records`)
+      if (iscnId) url.searchParams.append('iscn_id', iscnId)
+      if (owner) url.searchParams.append('owner', owner)
+      if (contentFingerprint) url.searchParams.append('fingerprint', contentFingerprint)
+      if (stakeholderId) url.searchParams.append('stakeholder.id', stakeholderId)
+      if (stakeholderName)url.searchParams.append('stakeholder.name', stakeholderName)
+      if (keyword) url.searchParams.append('keyword', keyword)
```
williamchong, line 164: Can we just use `qs.stringify(params)` here?

williamchong (changes requested):
How can I check for `stakeholderId` using wallet in this case?

pages/search/_keyword.vue
williamchong: I think single item is good enough for now

pages/view/_iscnId.vue
WeiJun0827: I think it's legacy code between version iteration

WeiJun0827:
Seems we cannot leave undefined params in indexer API:
- [x] [works](https://mainnet-node.like.co/iscn/records?keyword=DAO)
- [ ] [doesn't work](https://mainnet-node.like.co/iscn/records?iscn_id=&owner=&fingerprint=&stakeholder.id=&stakeholder.name=&keyword=DAO)

I still keep the conditional check before assigning params.

nwingt (approved):

PR #235, opened 2022-09-22
🐛 Don't show mint button in non popup view page

williamchong:
image
the button is very strange now, hide it for now


pages/view/_iscnId.vue
```diff
@@ -69,14 +69,6 @@
...
   </div>
   <div v-else>
-    <div v-if="isShowMintButton" class="flex justify-center w-full bg-like-cyan-light">
-      <Button
-        preset="plain"
-        :to="localeLocation({ name: 'nft-iscn-iscnId', params: { iscnId: iscnId } })"
-        class="text-like-green"
-        :text="$t('NFTPortal.button.preminted')"
-      />
-    </div>
```
nwingt, line 79:
```suggestion
    <div v-if="isShowMintButton" class="flex tail w-full">
      <Button
        preset="secondary"
        :to="localeLocation({ name: 'nft-iscn-iscnId', params: { iscnId: iscnId } })"
        class="text-like-green"
        :text="$t('NFTPortal.button.preminted')"
      />
    </div>
```

williamchong: So we prefer to keep this button?

nwingt (approved):

PR #236, opened 2022-09-23
👽️ Move keplr gas prices into fee currencies

williamchong:


nwingt (commented):
No need to update types?

williamchong:
> No need to update types?

We did not use keplr type here, just a placeholder `any`

williamchong:
Added keplr types

nwingt (approved):

PR #243, opened 2022-09-29
🐛 Fix if url encoded

Tsai-Hsueh-Kuan:
🐛 Fix if url encoded


WeiJun0827 (approved):

williamchong:
Modified into https://github.com/likecoin/app-like-co/pull/244

PR #270, opened 2022-12-07
🚸 <NFTMintProcess> Block minting with Wallet Connect

AuroraHuang22:
截圖 2022-12-07 上午10 41 00

### DIFF
--- a/pages/nft/iscn/_iscnId.vue
+++ b/pages/nft/iscn/_iscnId.vue
@@ -47,6 +47,7 @@
         :class-id="classId"
         :nft-link="apiData && detailsPageURL"
         :has-error="hasError"
+        :error-message="errorMessage"
         :tx-status="txStatus"
         :state="mintState"
       />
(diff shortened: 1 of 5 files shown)


nwingt:
Note sure if we want to block eariler

AuroraHuang22:
> Note sure if we want to block eariler, start from `/nft/url`

Also blocking Wallet Connect on nft/url

截圖 2022-12-07 下午5 36 43

pages/nft/iscn/_iscnId.vue
```diff
@@ -421,6 +421,10 @@ export default class NFTTestMintPage extends Vue {
         case 'create': {
           this.isLoading = true
           const isAllowed = IS_TESTNET || await this.checkIsWhitelisted();
+          if (this.isUsingLikerLandApp) {
+            this.errorMessage = this.$t('IscnRegisterForm.error.walletConnect') as string
+            throw new Error(ErrorType.USE_WALLET_CONNECT)
+          }
```
williamchong, line 427: How about move it before `switch (this.state) {`, after `balance` check?

nwingt (approved):

PR #281, opened 2023-01-05
🥅 Handle conflict error when posting mint info

AuroraHuang22:


pages/nft/iscn/_iscnId.vue
```diff
@@ -453,8 +456,11 @@ export default class NFTTestMintPage extends Vue {
...
-          await this.mintNFT()
-          if (!this.mintNFTResult) break
+          await this.checkIsNFTIdListExist()
+          if (!this.chainNFTIdList) {
+            await this.mintNFT()
+            if (!this.mintNFTResult) break
+          }
           if (this.isWritingNFT) {
             await this.postMintInfo()
             if (!this.postInfo) break
```
williamchong, line 466: Can we also just skip `this.mintNFT()` if `mintNFTResult` exists?
Using local state we can already know if minting is done, we only need to check chain API for new/refreshed sessions

williamchong (changes requested):

williamchong (approved):

PR #282, opened 2023-01-05
✨ Support passing custom NFT prefix via qs

williamchong:
image

### DIFF
--- a/pages/nft/iscn/_iscnId.vue
+++ b/pages/nft/iscn/_iscnId.vue
@@ -210,6 +210,10 @@ export default class NFTTestMintPage extends Vue {
     return !(nft && nft !== '0');
   }
 
+  get NFTPrefix(): string {
+    return (this.$route.query.nft_prefix as string) || 'Writing NFT';
+  }
+
   get hasOpener(): boolean {
     const { opener = 0 } = this.$route.query
     return window?.opener && (opener && opener !== '0');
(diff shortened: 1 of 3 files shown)


pages/nft/iscn/_iscnId.vue
```diff
@@ -336,7 +340,8 @@ export default class NFTTestMintPage extends Vue {
   }
 
   get NftName() {
-    return `${this.isWritingNFT ? 'Writing NFT - ' : ''}${this.iscnData?.contentMetadata?.name || 'NFT'}`;
+    const prefix = this.NFTPrefix || `${this.isWritingNFT ? 'Writing NFT' : ''}`;
```
nwingt, line 343: Seem `this.NFTPrefix` is used in this case??

nwingt (changes requested):

pages/view/_iscnId.vue
```diff
@@ -534,6 +534,15 @@ export default class ViewIscnIdPage extends Vue {
     return this.$route.query.layout === 'popup'
   }
 
+  get mintQuerys() {
```
nwingt, line 537:
```suggestion
  get mintQueries() {
```

pages/nft/iscn/_iscnId.vue
williamchong: Lets handle the isWritingNFT judgement in `this.NFTPrefix`

nwingt (approved):

PR #290, opened 2023-01-19
🚸 Hide close button in popup if mint NFT is next step

williamchong:


nwingt (approved):
Why there is a long and short button label?

williamchong:
> Why there is a long and short button label?

long button label is for the very long button (banner size) on the top of the page

PR #295, opened 2023-02-01
✨ Add mint button in homepage, block whitelist in nft-url

williamchong:
image
image
image


components/IscnInfoCard.vue
```diff
@@ -109,6 +109,28 @@
...
+        >
+          <Button
+            preset="secondary"
+            :text="$t('HomePage.mint.button')"
+            :to="localeLocation({ name: 'nft' })"
+            class="text-like-green"
+            :style="{ backgroundColor: '#AAF1E7', color: '#28646E' }"
+          >
+            <template #prepend>
+              <IconMint class="w-[20px]" />
```
nwingt, line 127: Please use this icon
mint-with-hex

components/IscnInfoCard.vue
```diff
@@ -109,6 +109,28 @@
...
+            'rounded-[20px]',
+            'my-[12px]'
+          ]"
+        >
+          <Button
+            preset="secondary"
+            :text="$t('HomePage.mint.button')"
+            :to="localeLocation({ name: 'nft' })"
+            class="text-like-green"
+            :style="{ backgroundColor: '#AAF1E7', color: '#28646E' }"
```
nwingt, line 124: Can this style be removed? I think it is secondary preset

components/IscnInfoCard.vue
```diff
@@ -109,6 +109,28 @@
             </template>
           </Button>
         </div>
+        <div
+          :class="[
+            'p-[8px]',
+            'rounded-[20px]',
```
nwingt, line 115:
```suggestion
```

components/IscnInfoCard.vue
nwingt: @williamchong Updated icon
mint-with-hex-updated

nwingt (approved):

PR #301, opened 2023-02-08
✨ Allow app to mint 35 wnft

williamchong:
Last time the limit we test is 40, but no idea how to test on rinkeby first before release though

### DIFF
--- a/pages/nft/iscn/_iscnId.vue
+++ b/pages/nft/iscn/_iscnId.vue
@@ -79,8 +79,7 @@
       </template>
     </ContentCard>
 
-    <AttentionsOpenLikerLandApp v-if="isUsingLikerLandApp" />
-    <AttentionsLedger v-else />
+    <AttentionsLedger />
     <AlertsSignFailed />
   </Page>
 </template>
@@ -384,6 +383,9 @@ export default class NFTTestMintPage extends Vue {
   }
 
   get premintAmount() {
+    if (this.isUsingLikerLandApp) {
+      return 35;
+    }
     return this.isWritingNFT ? 500 : 100;
   }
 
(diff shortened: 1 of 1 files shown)


pages/nft/iscn/_iscnId.vue
```diff
@@ -802,11 +800,20 @@ export default class NFTTestMintPage extends Vue {
...
+        let res;
+        res = await signingClient.sendMessages(this.address, mintMessages)
+        if (this.isWritingNFT) res = await signingClient.sendMessages(this.address, sendMessages)
+        this.mintNFTResult = res;
+      } else {
+        let messages = mintMessages;
+
+        if (this.isWritingNFT) messages = messages.concat(sendMessages);
+
+        logTrackerEvent(this, 'IscnMintNFT', 'MintNFT', this.classId, 1);
```
nwingt, line 814: Already logged above?

pages/nft/iscn/_iscnId.vue
```diff
@@ -802,11 +800,20 @@ export default class NFTTestMintPage extends Vue {
...
       )
-      let messages = mintMessages;
-      if (this.isWritingNFT) messages = messages.concat(sendMessages);
-
       logTrackerEvent(this, 'IscnMintNFT', 'MintNFT', this.classId, 1);
-      this.mintNFTResult = await signingClient.sendMessages(this.address, messages)
+      if (this.isUsingLikerLandApp) {
+        let res;
+        res = await signingClient.sendMessages(this.address, mintMessages)
+        if (this.isWritingNFT) res = await signingClient.sendMessages(this.address, sendMessages)
```
nwingt, line 807: So Liker Land app will sign twice, one for minting and transfer in WNFT flow?

williamchong: yes as to reduce the size
potentially it will cause recovery issue where user kept on minting nft but didn't send the nft away, we can add an extra state to handle that though

PR #306, opened 2023-02-09
🚸 Store pagination and use same window for work/search result

williamchong:
Store pagination in qs to allow backing to same search/work page state from view page

### DIFF
--- a/pages/search/_keyword.vue
+++ b/pages/search/_keyword.vue
@@ -38,7 +38,7 @@
           size="small"
           :text="$t('WorksPage.pagination.button.first')"
           :is-disabled="pageNumber === 0"
-          @click="pageNumber = 0"
+          @click="firstPage"
         />
         <Button
           class="text-dark-gray"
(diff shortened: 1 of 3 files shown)


AuroraHuang22 (approved):

pages/search/_keyword.vue
```diff
@@ -200,10 +201,30 @@ export default class SearchPage extends Vue {
...
+    this.$router.replace({
+      query: { ...this.$route.query, page: this.pageNumber.toString() },
+    });
+  }
+
+  lastPage() {
+    this.pageNumber = this.pages.length - 1
+    this.$router.replace({
+      query: { ...this.$route.query, page: this.pageNumber.toString() },
+    });
```
nwingt, line 227: I think using a watcher is more intuitive, keep repeating
```js
this.$router.replace({
  query: { ...this.$route.query, page: this.pageNumber.toString() },
});
```

pages/search/_keyword.vue
```diff
@@ -169,6 +170,13 @@ export default class SearchPage extends Vue {
     this.search()
   }
 
+  @Watch('page')
```
nwingt, line 173: Not `pageNumber`??

williamchong: Yeah brain fog... updated

nwingt (approved):

PR #312, opened 2023-02-15
💩 Mitigate SSR issue on view popup

williamchong:
I dont understand how store action in asyncData only cause issue under popup layout
but seems not fetching solves the issue


nwingt (approved):

williamchong:
@nwingt updated with new mitigation in `asyncData`

nwingt (approved):

PR #316, opened 2023-02-16
🚑️ Fix ISCN card not resizing correctly

williamchong:


nwingt (commented):
Why diff of indentation?

nwingt (approved):

PR #319, opened 2023-03-01
✨ Introduce @likecoin/wallet-connector for login management

AuroraHuang22:
temp_deploy to [app.testnet.like.co](https://app.testnet.like.co/)


utils/misc.ts
```diff
@@ -24,3 +24,12 @@ export function readImageType(buffer: ArrayBuffer) {
   if (buffer) return imageType(Buffer.from(buffer));
   return null;
 }
+
+export function catchAxiosError(promise: Promise<any>) {
```
williamchong, line 28: We do know the promise should be of axios http request type though?

constant/index.ts
```diff
@@ -88,3 +88,25 @@ export const LIKER_NFT_API_WALLET = IS_TESTNET ? 'like1yney2cqn5qdrlc50yr5l53898
...
+? 'https://node.testnet.like.co'
+: 'https://mainnet-node.like.co';
+
+export const LIKECOIN_CHAIN_DENOM = IS_TESTNET ? 'EKIL' : 'LIKE';
+
+export const LIKECOIN_CHAIN_MIN_DENOM = IS_TESTNET ? 'nanoekil' : 'nanolike';
+
+export const CIVIC_LIKER_V3_STAKING_ENDPOINT = IS_TESTNET
+  ? 'https://likecoin-public-testnet-5.netlify.app/validators'
+  : 'https://dao.like.co/validators';
```
williamchong, line 110: This has nothing to do with `CIVIC_LIKER_V3`, just url for staking

components/NFTMint/Process.vue
```diff
@@ -154,9 +154,10 @@ export default class UploadForm extends Vue {
       case 'sign':
         switch (this.walletType) {
           case 'keplr':
+          case 'keplr-mobile':
             return this.$t('NFTPortal.loadingMessage.keplr')
 
-          case 'likerland_app':
+          case 'liker-id':
             return this.$t('NFTPortal.loadingMessage.likerId')
```
williamchong, line 161: We don't support cosmos station here?

utils/logger.ts
```diff
@@ -24,6 +24,7 @@ export async function setLoggerUser(vue: Vue, { wallet } : { wallet?: string } =
       // https://support.google.com/analytics/answer/9973999?hl=en
       // vue.$gtag.config({ user_id: hashedId });
       vue.$gtag.set({ user_id: hashedId });
+      vue.$gtag.event('login', { method });
```
williamchong, line 27: Do we know if it's explicit login or just restore session? We don't want to call login event if it's just restored session

williamchong (changes requested):

utils/misc.ts
AuroraHuang22: Since [axios.get](https://github.com/likecoin/app-like-co/pull/319/files#diff-a44f2b879ce5cfedfee1de745a1ecc3eca574e5a45d4ae1040c80b846874c643R134-R135) itself returns an AxiosPromise, the catchAxiosError function is designed to handle the returned promise.

williamchong: Just type the param `promise` as `AxiosPromise` then?

components/RootLayout.vue
```diff
@@ -15,32 +15,19 @@
...
 
-    <DialogContainer>
-      <ConnectWalletDialog
-        :is-opened="isShowConnectWalletDialog"
-        @quit="handleConnectWalletDialogQuit"
-        @close="handleConnectWalletDialogClose"
-      />
-      <ConnectLikerIdDialog @quit="toggleConnectWalletDialog(true)" />
+<DialogContainer>
+      <Dialog />
```
nwingt, line 19: Incorrect indent?

layouts/wallet.vue
```diff
@@ -13,40 +13,58 @@
...
-
   get isHideFooter() {
     return this.$route.path.includes('/nft/purchase/') || this.$route.query.layout === 'popup';
   }
 
-  mounted() {
-    if (!this.currentAddress) {
-      this.toggleConnectWalletDialog(true)
+  async mounted() {
+    await this.restoreSession()
```
nwingt, line 39: Isn't `restoreSession()` already done in `<RootLayout/>`?

AuroraHuang22: Yes, but there is a time lag when refreshing the [mint /] or [publishing /] (RootLayout + WalletLayout) screens in place, causing the background to enter the RootLayout screen while the WalletLayout is still requesting login.

nwingt: I think `restoreSession` should set a state e.g. `isRestoringSession`??, so you don't double restoring the session

components/RootLayout.vue
```diff
@@ -56,7 +56,7 @@ export default class RootLayout extends Vue {
 
   async mounted() {
     this.initUIStore()
-    await this.restoreSession()
+    await this.restoreSessionIfNecessary()
```
nwingt, line 59: is it better to let `this.isOpenChainUpgradeBlockingDialog = !!IS_CHAIN_UPGRADING` execute first?

store/wallet.ts
```diff
@@ -93,222 +97,102 @@ export default class Wallet extends VuexModule {
...
-      this.context.commit('setSigner', offlineSigner);
-      this.context.commit('setAccounts', accounts);
-      window.localStorage?.setItem(KEY_CONNECTED_WALLET_TYPE, 'likerland_app');
-      return true
-    } catch (error) {
-      // eslint-disable-next-line no-console
-      console.error(error)
-    }
-    return false
+    return restoreSessionPromise;
```
nwingt, line 165: Need to reset the promise
```js
const result = await restoreSessionPromise;
restoreSessionPromise = null;
return result;
```

nwingt (approved):

PR #324, opened 2023-03-15
✨ Add nft subscription minting

williamchong:
For first release, no link to subscription page would be presented for common users
It is expected we would distribute such link by invite or via other channel
Subscription related status would only appear when user is a subscriber


pages/nft/iscn/_iscnId.vue
```diff
@@ -798,39 +845,52 @@ export default class NFTTestMintPage extends Vue {
 
   async createNftClass() {
     await this.initIfNecessary()
-    if (!this.signer) return
+    if (!this.signer) return '';
```
nwingt, line 848: Should this guard be put in the else below like L960?

nwingt (approved):

WeiJun0827 (approved):
Was thinking about whether we should handle the new subscription logic in `_iscnId.vue` with the strategy pattern. But maybe let it be for now?

williamchong:
> Was thinking about whether we should handle the new subscription logic in `_iscnId.vue` with the strategy pattern. But maybe let it be for now?

Seems to be a good idea, not sure how to implement in nuxt context through.
Do we need to pass `this` into constrcutor if we always call `this` in methods?

pages/nft/iscn/_iscnId.vue
williamchong: yes indeed

WeiJun0827:
> Seems to be a good idea, not sure how to implement in nuxt context through. Do we need to pass `this` into constrcutor if we always call `this` in methods?

Yeah, seems can do.

PR #329, opened 2023-03-20
🔥 Remove whitelist related checkers

AuroraHuang22:
網頁擷取_20-3-2023_221019_localhost


locales/en.json
```diff
@@ -261,6 +261,8 @@
   "NFTPortal.label.creator": "Creator",
   "NFTPortal.label.defaultDescription": "(no description)",
   "NFTPortal.label.detailsPage": "NFT Details Page :",
+  "NFTPortal.label.input.email": "Join our community and stay in the loop on exclusive product updates and offers. {link} now!",
+  "NFTPortal.label.input.link": "Please leave your email.",
```
williamchong, line 265:
```suggestion
  "NFTPortal.label.input.link": "Please leave your email",
```

components/MintPageContainer.vue
```diff
@@ -0,0 +1,75 @@
...
+              />
+            </template>
+          </i18n>
+        </Label>
+      </div>
+    </div>
+    <slot />
+    <div class="flex flex-col max-w-[600px]">
+      <div class="mt-[-20px]">
+        <AttentionsOpenLikerLandApp v-if="isUsingLikerLandApp" />
```
williamchong, line 52: We may want to refactor this to `isUsingWalletConnect` and support all wallet connect method in the future, since this component is used to remind people to open their app

pages/nft/history/_iscnId.vue
```diff
@@ -1,24 +1,13 @@
...
+  <MintPageContainer>
     <table>
       <tr v-for="r in records" :key="r.txHash">
         <td>Price: {{r.price}} LIKE </td>
         <td>Buyer: {{r.toWallet}} </td>
         <td>Time: {{new Date(r.timestamp)}}</td>
       </tr>
     </table>
-  </Page>
+  </MintPageContainer>
```
williamchong, line 10: I think we should remove this unused page..

pages/nft/purchase/_iscnId.vue
```diff
@@ -60,11 +60,7 @@
           </i18n>
         </Label>
       </Card>
-
-      <AttentionsOpenLikerLandApp v-if="isUsingLikerLandApp" />
-      <AttentionsLedger v-else />
     </div>
-    <AlertsSignFailed />
```
williamchong, line 67: I think we should remove this unused page..

components/MintPageContainer.vue
williamchong: Also `isStateTransaction` is removed, this prompt is actually less useful UX wise
Old behaviour is this component only appears when signing is needed, and is hidden otherwise. so people are reminded when to wait and when to open the app to continue
Can we implement this behaviour here somehow?

store/wallet.ts
```diff
@@ -43,6 +43,7 @@ export default class Wallet extends VuexModule {
   isOpenSnackbar = false
   likerInfo = null
   errorType = ''
+  hasEmail = false
```
williamchong, line 46:
```suggestion
  hasSubmittedEmail = false
```

store/wallet.ts
```diff
@@ -151,7 +167,6 @@ export default class Wallet extends VuexModule {
     if (session) {
       const { accounts, method } = session
       await this.context.dispatch('initWallet', { method, accounts })
-      await this.initWallet({ accounts, method })
```
williamchong, line 154: Why remove this line?

williamchong (changes requested):

store/wallet.ts
AuroraHuang22: Since we already have this.context.dispatch('initWallet', { method, accounts }), we probably don't need to repeat this.initWallet?

williamchong (approved):
Please confirm the final content and wording of the google form, I think we would also ask creator's website url

PR #339, opened 2023-04-20
🚸 Remove version from ISCN card QR code

nwingt:


components/IscnCard.vue
```diff
@@ -182,10 +182,21 @@ export default class IscnCard extends Vue {
     return `${type} ${timestamp} ${version ? 'Version ' : ''}${version}`
   }
 
+  get iscnIDParts() {
+    const parts = this.recordID.match(/(iscn:\/\/[a-zA-Z0-9-_]+\/([a-zA-Z0-9-_]+))\/\d+/)
```
nnkken, line 186: For an ISCN ID without version this won't match, so better make the version part optional
```suggestion
    const parts = this.recordID.match(/(iscn:\/\/[a-zA-Z0-9-_]+\/([a-zA-Z0-9-_]+))(?:\/\d+)?/)
```

nnkken: (May remove the `?:` to also capture the version if present)

williamchong (approved):

PR #342, opened 2023-04-26
🚸 Replace the standard strings with actual error codes

AuroraHuang22:


williamchong (commented):
Should we call err.toString() instead?

nwingt (approved):
Debug first

PR #349, opened 2023-05-23
✨ Use fingerprint as og image if ISCN type is Photo

williamchong:


AuroraHuang22 (approved):

williamchong:
update: reset arweaveid on image change

WeiJun0827 (approved):

PR #353, opened 2023-06-01
⚡️ Merge new nft class and create royalty tx

williamchong:
Pre-calculate class id when creating class + royalty config
next step might be merging create class, royalty config, mint and send into 1 tx

### DIFF
--- a/pages/nft/iscn/_iscnId.vue
+++ b/pages/nft/iscn/_iscnId.vue
@@ -111,16 +111,20 @@
 <script lang="ts">
 import qs from 'querystring'
 // eslint-disable-next-line import/no-extraneous-dependencies
+import Long from 'long';
+// eslint-disable-next-line import/no-extraneous-dependencies
 import { OfflineSigner } from '@cosmjs/proto-signing'
 import { parseAndCalculateStakeholderRewards } from '@likecoin/iscn-js/dist/iscn/parsing';
 import { Vue, Component } from 'vue-property-decorator'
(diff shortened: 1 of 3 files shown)


pages/nft/iscn/_iscnId.vue
```diff
@@ -898,20 +902,24 @@ export default class NFTTestMintPage extends Vue {
         await signingClient.setSigner(this.signer)
         logTrackerEvent(this, 'IscnMintNFT', 'CreateNftClass', this.iscnId, 1);
         this.txStatus = TxStatus.PROCESSING
-        const res = await signingClient.createNFTClass(
+
+        const queryClient = signingClient.getISCNQueryClient()
+        const q = await queryClient.getQueryClient()
+        const { classes = [] } = await q.likenft.classesByISCN(this.iscnId).catch(() => ({} as any))
+        const existingClassesLength = classes?.length || 0;
```
nnkken, line 909: Any pagination on this API?

williamchong: Yes but for this case we can safely assume <100 so no pages
It is expected there is no existing NFT under the ISCN, if there is we are probably just handling 1-2 wrongly minted classes

nnkken (approved):

pages/nft/iscn/_iscnId.vue
williamchong: added comment for this

PR #378, opened 2023-07-31
[TAS-134] ✏️ Fix typos

WeiJun0827:


williamchong:
the existing metadata would be kept as is on chain even after this fix
need to handle both key on frontend i guess

PR #382, opened 2023-08-15
[TAS-205] ✨ Add 'Collectable Before' feature

WeiJun0827:
Depends on https://github.com/likecoin/likecoin-api-public/pull/506

https://github.com/likecoin/app-like-co/assets/33746295/5e900318-6728-4ec5-b7a6-3e22bf9ae57c

### DIFF
--- a/components/NFTMint/WriterMessage.vue
+++ b/components/NFTMint/WriterMessage.vue
@@ -148,15 +148,15 @@
             </option>
           </select>
         </div>
-      </div>
-      <div
-        class="flex flex-col justify-start gap-[32px] p-[20px] pb-[24px] border-2 border-[#E6F4F2] rounded-[16px] w-full"
-      >
-        <label>
-          <input v-model="shouldShowCollectExpiryDateInput" type="checkbox"  @change="resetCollectExpiryDate" />
-          {{ $t('NFTPortal.label.collectExpiryDate.input') }}
-        </label>
-        <input v-if=shouldShowCollectExpiryDateInput ref="collectExpiryDateInput" type="date" :min="tomorrow" @change="updateCollectExpiryDate($event)" />
+        <div
+          class="flex justify-between gap-[12px] text-dark-gray text-[14px] items-center"
+        >
+          <label>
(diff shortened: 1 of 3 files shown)


locales/en.json
```diff
@@ -263,6 +263,7 @@
   "NFTPortal.errorMessage.metadata.noURL": "Warning: No URL in ISCN's metadata {more}.",
   "NFTPortal.errorMessage.metadata.noURL.learnMore": "learn more",
   "NFTPortal.label.classId": "Class ID",
+  "NFTPortal.label.collectableBefore.input": "Set collectable expiration date",
```
williamchong, line 266:
```suggestion
  "NFTPortal.label.collectableBefore.input": "Set a deadline for collection",
```

components/NFTMint/WriterMessage.vue
```diff
@@ -202,6 +217,20 @@ export default class WriterMessage extends Vue {
     this.$emit('update-mint-amount', mintAmount)
   }
 
+  resetCollectableBeforeDate(value: Event) {
+    const { checked } = (value.target as HTMLInputElement)
+    if (!checked) {
+      this.$emit('update:collectableBeforeDate', '')
+      this.$emit('update-collectable-before-date', '')
+    }
```
williamchong, line 225: Would this break if I uncheck then check the checkbox? A date input would be shown, but no `collectable-before-date` event is fired since no @change has occurred yet

williamchong (changes requested):

components/NFTMint/WriterMessage.vue
WeiJun0827: The field would be reset if the checkbox is unchecked:

https://github.com/likecoin/app-like-co/assets/33746295/b76aa335-73ff-442f-8a38-fc1043cf680f

Is this behavior works OK?

WeiJun0827: The above actions didn't click the background actually, I think the `computed` did the job, and the Vue extension UI has a little latency.

williamchong (approved):

WeiJun0827:
@williamchong  Updated
截圖 2023-08-21 下午5 45 31

williamchong (approved):

PR #385, opened 2023-09-07
[TAS-114] ✨ Add upload only mode

williamchong:
image
image


components/IscnRegisterForm.vue
```diff
@@ -308,16 +334,35 @@
...
+                  'text-[8px]',
+                  'underline',
+                  'mt-[4px]',
+                  'mb-[-12px]',
+                ]"
+                href="#"
+                @click.prevent="showUploadOnly = true"
+              >
+                {{ $t('IscnRegisterForm.button.uploadOnly') }}
+              </a>
```
nwingt, line 364: I believe this button should be moved to the left, as the main action should always remain on the right-most side

Also, this `<a/>` could be replaced by `<Button preset="outline" />` to ensure consistent styling

components/IscnRegisterForm.vue
```diff
@@ -308,16 +334,35 @@
...
-            <Button
-              v-else
-              :text="$t('IscnRegisterForm.button.register')"
-              type="submit"
-              preset="secondary"
+            <div v-else
+              :class="[
+                'flex',
+                'flex-col',
+                'items-end',
```
nwingt, line 341:
```suggestion
                'items-end',
                'gap-[12px]',
```

components/IscnRegisterForm.vue
```diff
@@ -1020,6 +1068,33 @@ export default class IscnRegisterForm extends Vue {
     return undefined
   }
 
+  async onUploadOnly(): Promise<void> {
+    logTrackerEvent(this, 'ISCNCreate', 'ClickSubmit', '', 1);
```
nwingt, line 1072: Should we distinguish this from the `ClickSubmit` event in `onSubmit()`?

components/FileUploadedInfo.vue
```diff
@@ -0,0 +1,63 @@
...
+      >
+        <a
+          :href="`https://arweave.net/${arweaveId}`"
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          {{ arweaveId }}
+        </a>
+      </FormField>
+      <Divider class="mb-[24px]" />
```
nwingt, line 49:
```suggestion
      <Divider v-if="$slots.footer" class="mb-[24px]" />
```

components/FileUploadedInfo.vue
```diff
@@ -0,0 +1,63 @@
...
+      <slot name="footer" />
```
nwingt, line 50: Seems this is not used

components/FileUploadedInfo.vue
```diff
@@ -0,0 +1,63 @@
...
+      :has-padding="false"
+    >
+      <!-- header -->
+      <IscnFormHeader :step="step" :total-step="4"/>
+      <!-- guide title -->
+      <Label
+        :text="$t('FileUploaded.guide.title')"
+        class="text-medium-gray mt-[12px] mb-[28px]"
+      />
+      <slot name="card-footer" />
```
nwingt, line 17: Seems this is not used

components/IscnRegisterForm.vue
williamchong: > I believe this button should be moved to the left, as the main action should always remain on the right-most side
> 
> Also, this `<a/>` could be replaced by `<Button preset="outline" />` to ensure consistent styling

Totally missed there are 2 parts of review..
image

Does this look ok?

PR #388, opened 2023-09-19
[TAS-376]🎨 Change initial price to USD

AuroraHuang22:
截圖 2023-09-19 下午11 31 54


williamchong (approved):
We might need more tiers for usd

williamchong (approved):

WeiJun0827:
Can we also update the default `initialBatch` to 7?
https://github.com/likecoin/app-like-co/blob/main/pages/nft/iscn/_iscnId.vue#L266

WeiJun0827 (approved):

PR #404, opened 2023-10-30
[TAS-485] 💄 Wrap stakeholder tags

AuroraHuang22:
截圖 2023-10-30 上午9 20 50

### DIFF
--- a/pages/view/_iscnId.vue
+++ b/pages/view/_iscnId.vue
@@ -260,13 +260,12 @@
           </template>
           <FormField
             :label="$t('iscn.meta.stakeholders')"
-            class="mb-[6px]"
-            content-classes="flex flex-row flex-wrap items-center"
+            class="mb-[12px]"
+            content-classes="flex flex-row flex-wrap items-center gap-[8px]"
           >
             <Button
               v-for="(stakeholder, index) in stakeholders"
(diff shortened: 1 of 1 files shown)


pages/view/_iscnId.vue
```diff
@@ -260,13 +260,13 @@
...
-            class="mb-[12px]"
-            content-classes="flex flex-row items-center"
+            class="mb-[6px]"
+            content-classes="flex flex-row flex-wrap items-center"
           >
             <Button
               v-for="(stakeholder, index) in stakeholders"
               :key="stakeholder.key"
-              class="mr-[8px]"
+              class="mr-[8px] mb-[8px]"
```
nwingt, line 269: Better use `gap-[8px]` and remove all the margin

nwingt (changes requested):

pages/view/_iscnId.vue
```diff
@@ -260,13 +260,12 @@
           </template>
           <FormField
             :label="$t('iscn.meta.stakeholders')"
-            class="mb-[12px]"
-            content-classes="flex flex-row items-center"
+            class="mb-[6px]"
```
nwingt, line 263: Should we keep `mb-[12px]`?

PR #405, opened 2023-10-30
[TAS-597] 🚚 Separate upload only mode and register ISCN

AuroraHuang22:
https://github.com/likecoin/app-like-co/assets/75730405/36bfc1aa-0be7-45a5-9047-3028600b9beb

### DIFF
--- a/components/IscnUploadForm.vue
+++ b/components/IscnUploadForm.vue
@@ -484,6 +484,8 @@ export default class UploadForm extends Vue {
   onOpenKeplr() {
     logTrackerEvent(this, 'ISCNCreate', 'OpenKeplr', '', 1);
     this.isOpenKeplr = true
+    // Hack: In some cases, there might be no response from Keplr,
+    // so we set a timeout to automatically close it after 5 seconds.
     setTimeout(() => {
       this.isOpenKeplr = false
     }, 5000)
--- a/components/IscnUploadForm.vue
+++ b/components/IscnUploadForm.vue
@@ -308,10 +308,13 @@ export default class UploadForm extends Vue {
 
   uploadSizeLimit: number = UPLOAD_FILESIZE_MAX
   uploadStatus: string = ''
-  uploadArweaveIdList: string[] = []
   arweaveFeeTargetAddress: string = ''
   arweaveFee = new BigNumber(0)
-  sentArweaveTransactionHashes = new Map<string, string>()
+  sentArweaveTransactionHashes = new Map<
+    string, { transactionHash?: string, arweaveId?: string }
+  >()
+
+
 
   likerId: string = ''
   likerIdsAddresses: (string | void)[] = []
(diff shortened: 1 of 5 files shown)


components/IscnUploadForm.vue
```diff
@@ -190,9 +301,29 @@ export default class UploadForm extends Vue {
...
+  arweaveFeeTargetAddress: string = ''
+  arweaveFee = new BigNumber(0)
+  sentArweaveTransactionHashes = new Map<
+    string, { transactionHash?: string, arweaveId?: string }
+  >()
+
+
+
+  likerId: string = ''
+  likerIdsAddresses: (string | void)[] = []
```
williamchong, line 320: in what case would this be `void[]`?

components/IscnUploadForm.vue
```diff
@@ -334,5 +516,171 @@ export default class UploadForm extends Vue {
     this.displayImageSrc = this.fileRecords[index].fileData
     this.displayExifInfo = this.fileRecords[index].exifInfo
   }
+
+  onOpenKeplr() {
+    logTrackerEvent(this, 'ISCNCreate', 'OpenKeplr', '', 1);
+    this.isOpenKeplr = true
+    setTimeout(() => {
+      this.isOpenKeplr = false
+    }, 5000)
```
williamchong, line 525: Seems this is old code, but what does this magic 5s do?

components/IscnUploadForm.vue
```diff
@@ -334,5 +516,171 @@ export default class UploadForm extends Vue {
...
+  async estimateArweaveFee(): Promise<void> {
+    try {
+      const pricePromises = this.fileRecords.map(record =>
+        estimateBundlrFilePrice({
+          fileSize: record.fileBlob?.size || 0,
+          ipfsHash: record.ipfsHash,
+        }).then(priceResult => ({
+          ...priceResult,
+          ipfsHash: record.ipfsHash,
+        })),
```
williamchong, line 537: prefer `await` instead of `.then()` if you don't need to `.catch()` individual errors

components/IscnUploadForm.vue
```diff
@@ -334,5 +516,171 @@ export default class UploadForm extends Vue {
...
+
+  async onSubmit() {
+    if (IS_CHAIN_UPGRADING) return
+    logTrackerEvent(this, 'ISCNCreate', 'ClickUpload', '', 1);
+    this.uploadStatus = 'uploading'
+    this.$emit('handleSubmit')
+    this.error = ''
+    this.signDialogError = ''
+
+    const [balance] = await Promise.all([getAccountBalance(this.address)])
```
williamchong, line 648: Remove `Promise.all` for single promise

components/IscnUploadForm.vue
```diff
@@ -190,9 +301,29 @@ export default class UploadForm extends Vue {
...
+  isOpenSignDialog = false
+  isOpenWarningSnackbar = false
+  isOpenKeplr = true
   isSizeExceeded = false
 
   uploadSizeLimit: number = UPLOAD_FILESIZE_MAX
+  uploadStatus: string = ''
+  arweaveFeeTargetAddress: string = ''
+  arweaveFee = new BigNumber(0)
+  sentArweaveTransactionHashes = new Map<
```
williamchong, line 313: Consider renaming to `sentArweaveTransactionInfo` instead of hashes

williamchong (changes requested):
Better split the commits when you are doing both moving code and writing new logic. That way we can skip through commit that only moves code quickly, and focus on reviewing new logic.

components/IscnUploadForm.vue
AuroraHuang22: `likerIdsAddresses` was originally directly awaiting the result of a Promise.all

components/IscnUploadForm.vue
AuroraHuang22: sometimes encountering no response from Keplr when calling `isOpenKeplr = true,` so we added a setTimeout to restart it.

AuroraHuang22: But I don't know why it works.🌚

williamchong: Maybe add a comment here to explain

williamchong (approved):

PR #408, opened 2023-11-03
[TAS-584] 🚸 Extract Epub metadata for ISCN registration

AuroraHuang22:


locales/en.json
```diff
@@ -238,6 +240,7 @@
   "IscnRegisterForm.placeholder.fileName": "File Name (en)",
   "IscnRegisterForm.placeholder.fileType": "File Type",
   "IscnRegisterForm.placeholder.url": "URL",
+  "IscnRegisterForm.placeholder.isbn": "ISBN 10/ ISCN 13",
```
williamchong, line 243:
```suggestion
  "IscnRegisterForm.placeholder.isbn": "ISBN 10/ ISBN 13",
```

utils/cosmos/iscn/iscn.type.ts
```diff
@@ -24,6 +24,9 @@ export interface ISCNRegisterPayload {
   stakeholders?: any[];
   recordNotes?: string;
   memo?: string;
+  inLanguage?: string;
+  thumbnailUrl?: string;
+  isbn?: string | number;
```
williamchong, line 29: Can we just cast all as either string or number? having both is not good for parsing

components/IscnUploadForm.vue
```diff
@@ -451,6 +456,106 @@ export default class UploadForm extends Vue {
     }
   }
 
+  async processEPub({ buffer, file }: { buffer: ArrayBuffer; file: File }) {
+    try {
+      const Book = ePub(buffer)
```
williamchong, line 461: prefer lowercap naming (`book` or `bookInstance`) for object instances, capital name are for classes

components/IscnUploadForm.vue
```diff
@@ -451,6 +456,106 @@ export default class UploadForm extends Vue {
...
+      const opfContent = await Book.archive.getText(opfFilePath)
+      const parser = new DOMParser()
+      const opfDocument = parser.parseFromString(opfContent, 'application/xml')
+      const dcSubjectElements = opfDocument.querySelectorAll(
+        'dc\\:subject, subject',
+      )
+      const subjects: string[] = []
+      dcSubjectElements.forEach((element) => {
+        const subject = element.textContent
+        subject && subjects.push(subject)
```
williamchong, line 486: Prefer explicit `if`

components/IscnUploadForm.vue
```diff
@@ -451,6 +456,106 @@ export default class UploadForm extends Vue {
...
+      if (metadata) {
+        epubMetadata.epubFileName = file.name
+        epubMetadata.title = metadata.title
+        epubMetadata.author = metadata.creator
+        epubMetadata.language = this.formatLanguage(metadata.language)
+        epubMetadata.description = metadata.description
+      }
+
+      // Get tags
+      const opfFilePath = await (Book.path as any).path
```
williamchong, line 476: Can we cast this as `Path` from epub.js?

components/IscnUploadForm.vue
```diff
@@ -451,6 +456,106 @@ export default class UploadForm extends Vue {
...
+      )
+      const subjects: string[] = []
+      dcSubjectElements.forEach((element) => {
+        const subject = element.textContent
+        subject && subjects.push(subject)
+      })
+      epubMetadata.tags = subjects
+
+      // Get cover file
+      const coverUrl = (Book as any).cover
```
williamchong, line 491: Can we use documented method `[coverUrl](https://github.com/futurepress/epub.js/blob/master/documentation/md/API.md#coverurl)` instead?

components/IscnUploadForm.vue
```diff
@@ -451,6 +456,106 @@ export default class UploadForm extends Vue {
...
+      if (!coverUrl) {
+        this.epubMetadataList.push(epubMetadata)
+        return
+      }
```
williamchong, line 495: better to use `if (!coverUrl) {} else {}` or `if (coverUrl) {} else {}` instead of early return here, since people may add more code down below without realizing it is actually an if else case if we use early `return`

components/IscnUploadForm.vue
```diff
@@ -451,6 +456,106 @@ export default class UploadForm extends Vue {
...
+          reader.onload = (e) => {
+            if (!e.target) return
+            fileRecord.fileData = e.target.result as string
+          }
+          reader.readAsDataURL(coverFile)
+          this.epubMetadataList = [
+            ...this.epubMetadataList,
+            epubMetadata,
+          ]
+          this.fileRecords.push(fileRecord)
```
williamchong, line 540: Should wrap this `onload` function in await, dangerous use of race condition here
or a simpler way is if we don't need `fileRecord` if `fileRecord.fileData` not exists, we should put `this.fileRecords.push(fileRecord)` inside `onload`

components/IscnUploadForm.vue
```diff
@@ -451,6 +456,106 @@ export default class UploadForm extends Vue {
...
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  formatLanguage(language: string) {
+    if (language && language.toLowerCase().startsWith('en')) {
+      return 'en'
+    }
+    if (language && language.toLowerCase().startsWith('zh')) {
+      return 'zh'
+    }
```
williamchong, line 555:
```suggestion
    if (language) {
      if (language.toLowerCase().startsWith('en')) {
        return 'en'
      } else if (language.toLowerCase().startsWith('zh')) {
        return 'zh'
      }
    }
```

components/IscnUploadForm.vue
```diff
@@ -472,7 +577,13 @@ export default class UploadForm extends Vue {
   }
 
   handleDeleteFile(index: number) {
-    this.fileRecords.splice(index, 1)
+    const deletedFile = this.fileRecords[index];
+    this.fileRecords.splice(index, 1);
+
+    const indexToDelete = this.epubMetadataList.findIndex(item => item.epubFileName === deletedFile.fileName);
```
williamchong, line 583: should we assign unique random id to these items instead of using `fileName`? `fileName` might not be unique

components/SameAsFieldList.vue
```diff
@@ -159,5 +174,34 @@ export default class WalletFieldList extends Vue {
...
+        case fileType.includes('jpg'):
+        case fileType.includes('jpeg'):
+          formattedFileType = 'jpg'
+          break
+        case fileType.includes('png'):
+          formattedFileType = 'png'
+          break
+        case fileType.includes('audio'):
+          formattedFileType = 'mp3'
+          break
```
williamchong, line 192: Why .mp3? We produce .ogg audio from our TTS tool for instance

components/IscnUploadForm.vue
```diff
@@ -637,12 +754,26 @@ export default class UploadForm extends Vue {
...
 
     const uploadArweaveIdList = Array.from(this.sentArweaveTransactionInfo.values()).map(entry => entry.arweaveId);
-    this.$emit('submit', { fileRecords: this.fileRecords, arweaveIds: uploadArweaveIdList })
+    this.fileRecords.forEach((record: any, index:number) => {
+      if (this.sentArweaveTransactionInfo.has(record.ipfsHash)) {
+        const arweaveId = this.sentArweaveTransactionInfo.get(
+          record.ipfsHash,
+        )?.arweaveId
+        if (arweaveId) {
+          this.fileRecords[index].arweaveId = `ar://${arweaveId}`
```
williamchong, line 772: having `.arweaveId` prefixed with `ar://` seems like a bad idea, what is the usecase here?

williamchong (changes requested):

components/IscnUploadForm.vue
AuroraHuang22: I think their type `Path` is outdated; the data I'm getting does not match the type they [originally set](https://github.com/futurepress/epub.js/blob/master/types/utils/path.d.ts).
截圖 2023-11-06 上午8 46 37

components/IscnUploadForm.vue
```diff
@@ -490,57 +490,59 @@ export default class UploadForm extends Vue {
...
           }
-          reader.readAsDataURL(coverFile)
-          this.epubMetadataList = [
-            ...this.epubMetadataList,
-            epubMetadata,
-          ]
-          this.fileRecords.push(fileRecord)
         }
+      } else {
+        this.epubMetadataList.push(epubMetadata)
```
williamchong, line 545: Do we need `this.epubMetadataList.push(epubMetadata)` also for `if (coverUrl)`? Maybe no need for `else`?

AuroraHuang22: The epubMetadata also includes the [title, description, and tags,](https://github.com/likecoin/app-like-co/pull/408/files#diff-b6c5216e116e745da273c1edb6dcccaa190f6d360b269cf1ebe0119d3c6d66d0R466-R473) so it needs to be saved even without a coverUrl.

williamchong: Need to remove `else` then

williamchong (approved):

williamchong (approved):

PR #440, opened 2024-02-15
[TAS-1217] 🚸 Add default value to stakeholders

AuroraHuang22:
截圖 2024-02-15 下午5 11 51

### DIFF
--- a/components/IscnRegisterForm.vue
+++ b/components/IscnRegisterForm.vue
@@ -1131,31 +1131,35 @@ export default class IscnRegisterForm extends Vue {
     }
 
     if (this.address) {
-      try {
-        const { data } = await this.$axios.get(
-          getUserInfoMinByAddress(this.address),
-        )
-        const iscnOwner = {
-          name: data?.displayName || this.address,
-          wallet: [{
-              content: this.address,
-              id: 1,
-              type: 'like',
-              isOpenOptions: false,
-            }],
-          url: [],
-          likerId: data?.user || '',
-          authorDescription: data?.description || 'Publisher',
-        }
-        this.authors.push(iscnOwner)
-      } catch (error) {
-        // eslint-disable-next-line no-console
-        console.error(error)
-      }
+      const iscnOwner = await this.fetchUserInfoByAddress(this.address)
+      this.authors.push(iscnOwner)
     }
     // ISCN Fee needs Arweave fee to calculate
     await this.calculateISCNFee()
     this.uploadStatus = ''
+  }
+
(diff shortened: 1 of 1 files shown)


components/IscnRegisterForm.vue
```diff
@@ -1114,38 +1115,48 @@ export default class IscnRegisterForm extends Vue {
...
-      }],
-      url: [],
-      likerId: '',
-      authorDescription: 'ISCN owner',
+
+    if (this.address) {
+      try {
+        const { data } = await this.$axios.get(
+          getUserInfoMinByAddress(this.address),
+        )
```
williamchong, line 1137: Will this throw 404 error, skipping the whole `iscnOwner` block if the address does not have likeId?

williamchong (changes requested):

components/IscnRegisterForm.vue
AuroraHuang22: [🎨 Use default info when likerid not found](https://github.com/likecoin/app-like-co/pull/440/commits/1f52588176c140fb6d65b6939d6182a5ea3228d0)

components/IscnRegisterForm.vue
```diff
@@ -1114,38 +1115,52 @@ export default class IscnRegisterForm extends Vue {
...
+      // eslint-disable-next-line no-console
+      console.error(error)
+      return {
+        name: address,
+        wallet: [{ content: address, id: 1, type: 'like', isOpenOptions: false }],
+        url: [],
+        likerId: '',
+        authorDescription: 'Publisher',
+      }
     }
```
williamchong, line 1162:
```suggestion
    let userData: any = null;
    try {
      ({ data: userData } = await this.$axios.get(getUserInfoMinByAddress(address)))
    } catch (error) {
      // eslint-disable-next-line no-console
      console.error(error)
    }
    return {
      name: userData?.displayName || address,
      wallet: [{ content: address, id: 1, type: 'like', isOpenOptions: false }],
      url: [],
      likerId: userData?.user || '',
      authorDescription: userData?.description || 'Publisher',
    }
```

williamchong (changes requested):
Also we can consider using `catchAxiosError`, but I think even non-404 error should be caught in this case

components/IscnRegisterForm.vue
AuroraHuang22: [🎨 Handle default info when user data is unavailable](https://github.com/likecoin/app-like-co/pull/440/commits/4474a883bc8b1d79c1d50380cd17c25a05fdfcfa)

PR #447, opened 2024-02-21
🐛 Use iscn prefix for update iscn msg

williamchong:


AuroraHuang22 (approved):

nwingt (approved):

williamchong:
Turns out this would break since somehow we require explicit update when updating...

PR #452, opened 2024-03-07
✨ Add authcore login

williamchong:
Also changes some likerland app and keplr hardcode condition/text


pages/auth/redirect.vue
```diff
@@ -0,0 +1,73 @@
...
+      } catch (err) {
+        this.$nuxt.error({
+          statusCode: 400,
+          message: (err as Error).toString(),
+        });
+      }
+    } else {
+      if (window.sessionStorage) {
+        window.sessionStorage.removeItem('USER_POST_AUTH_ROUTE');
+      }
```
AuroraHuang22, line 65: Here are some duplicated actions, maybe we could reorganize them like this:
```
manageSessionStorage(key: string, value?: string): void {
  if (window.sessionStorage) {
    if (value !== undefined) {
      window.sessionStorage.setItem(key, value);
    } else {
      window.sessionStorage.removeItem(key);
    }
  }
}

const USER_POST_AUTH_ROUTE = 'USER_POST_AUTH_ROUTE'

manageSessionStorage(USER_POST_AUTH_ROUTE, storedRoute); 
manageSessionStorage(USER_POST_AUTH_ROUTE);
```

AuroraHuang22 (approved):

pages/auth/redirect.vue
williamchong: Maybe when we reach rule of three

PR #456, opened 2024-03-18
[TAS-1386] 🔥 Remove content width settings

AuroraHuang22:
Remove the mode button that changes the content width so that the UI can be displayed based on the current window size.


layouts/default.vue
```diff
@@ -29,44 +29,14 @@
...
 import { Vue, Component } from 'vue-property-decorator'
-import { namespace } from 'vuex-class'
-
-const uiModule = namespace('ui')
 
 @Component
-export default class DefaultLayout extends Vue {
-  @uiModule.Getter isDesktopViewMode!: boolean
-  @uiModule.Action enableDesktopViewMode!: () => void
-  @uiModule.Action enableMobileViewMode!: () => void
```
williamchong, line 55: Remove these in store also, if not used any where else?

williamchong (approved):

PR #463, opened 2024-04-25
[TAS-1271] ✨ Add pdf support for iscn page

williamchong:
image
TODO: Fix strange full width number rendering for noto sans tc under pdf-lib ...


components/IscnUploadForm.vue
```diff
@@ -396,14 +400,19 @@ export default class IscnUploadForm extends Vue {
   }
 
   get showAddISCNPageOption() {
-    return this.mode === MODE.EDIT && this.fileRecords.some(file => file.fileType === 'application/epub+zip')
+    return this.mode === MODE.EDIT
+      && this.fileRecords.some(file => file.fileType === 'application/epub+zip'
+        || file.fileType === 'application/pdf')
```
nwingt, line 405: Maybe we just pick the one in line 411
```js
file.fileType === 'application/epub+zip' || file.fileType === 'application/pdf'
```
```js
['application/epub+zip', 'application/pdf'].includes(file.fileType)
```

PR #471, opened 2024-08-27
[TAS-2030] 💄 Add inLanguage field for book

AuroraHuang22:
截圖 2024-08-27 下午2 48 49

### DIFF
--- a/components/IscnRegisterForm.vue
+++ b/components/IscnRegisterForm.vue
@@ -140,7 +140,7 @@
           />
         </FormField>
       </div>
-      
+
       <Divider class="my-[12px]" />
       <!-- form fieldset -->
       <div>
@@ -867,7 +867,14 @@ export default class IscnRegisterForm extends Vue {
   }
 
   get defaultLanguage() {
-    return this.epubMetadata?.language || 'zh'
+    const containsChinese = (str: any) => /[\u4E00-\u9FFF\u3400-\u4DBF\uF900-\uFAFF]/.test(str);
+
+    const isEpubInChinese = this.epubMetadata?.language === 'zh'
+
+    const titleContainsChinese = this.name ? containsChinese(this.name) : isEpubInChinese;
+    const descriptionContainsChinese = this.description ? containsChinese(this.description) : isEpubInChinese;
+
+    return (titleContainsChinese || descriptionContainsChinese) ? 'zh' : 'en';
   }
 
   get authorDialogTitle() {
@@ -1087,6 +1094,11 @@ export default class IscnRegisterForm extends Vue {
     if (errormsg) this.isOpenWarningSnackbar = true
   }
 
+  @Watch('defaultLanguage', { immediate: true })
+  setLanguageValue(value: any) {
+    this.language = value
+  }
+
   async mounted() {
     this.uploadStatus = 'loading'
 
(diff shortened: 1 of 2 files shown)


components/IscnRegisterForm.vue
```diff
@@ -850,6 +866,10 @@ export default class IscnRegisterForm extends Vue {
     return 'CreativeWork'
   }
 
+  get defaultLanguage() {
+    return this.epubMetadata?.language || 'zh'
+  }
```
williamchong, line 871: Should set `this.language` as `defaultLanguage` on `mounted()`

williamchong: Also would be nice if we can set default as `en` or `zh` according to if we have non-ascii charater in `title` or `description`

williamchong (changes requested):

components/IscnRegisterForm.vue
AuroraHuang22: Use @watch to detect name & description language for book.

[🚸 Auto detect language for book](https://github.com/likecoin/app-like-co/pull/471/commits/6af5f50ddf8443855607eca376d6ba4e8c009f99)

components/IscnRegisterForm.vue
```diff
@@ -850,6 +866,17 @@ export default class IscnRegisterForm extends Vue {
     return 'CreativeWork'
   }
 
+  get defaultLanguage() {
+    const containsChinese = (str: any) => /[\u4E00-\u9FFF\u3400-\u4DBF\uF900-\uFAFF]/.test(str);
```
williamchong, line 870: Suggest to just simply check for all ascii/alphanumeric for English and avoid unicode logic

components/IscnRegisterForm.vue
```diff
@@ -850,6 +866,17 @@ export default class IscnRegisterForm extends Vue {
...
+
+    const isEpubInChinese = this.epubMetadata?.language === 'zh'
+
+    const titleContainsChinese = this.name ? containsChinese(this.name) : isEpubInChinese;
+    const descriptionContainsChinese = this.description ? containsChinese(this.description) : isEpubInChinese;
+
+    return (titleContainsChinese || descriptionContainsChinese) ? 'zh' : 'en';
+  }
```
williamchong, line 878: Cleaner to just at the end `return (isEpubInChinese  || titleContainsChinese || descriptionContainsChinese) ? 'zh' : 'en'` and don't use `isEpubInChinese` in 874 and 875

williamchong (changes requested):

components/IscnRegisterForm.vue
AuroraHuang22: [🎨 Simplify default language check](https://github.com/likecoin/app-like-co/pull/471/commits/7013a40c386ad473a0d62cd0b2a72b6d17512961)

components/IscnRegisterForm.vue
```diff
@@ -850,6 +866,17 @@ export default class IscnRegisterForm extends Vue {
     return 'CreativeWork'
   }
 
+  get defaultLanguage() {
+    const containsOnlyASCII = (str: string) => /^[ -~]*$/.test(str);
+
+    const isEpubInChinese = this.epubMetadata?.language === 'zh';
```
williamchong, line 872: Should we respect if epub metadata language is explicitly set to en?

AuroraHuang22: [🎨 Prioritize language from epubMetadata](https://github.com/likecoin/app-like-co/pull/471/commits/7f5a5a736db1c90c64104b9d9de553a2262d92d4)

williamchong (approved):

nwingt (approved):

PR #494, opened 2024-10-28
[TAS-1908] 👽️ Add auth to arweave upload api

williamchong:
https://github.com/likecoin/likecoin-api-public/pull/800


components/IscnUploadForm.vue
```diff
@@ -286,6 +287,7 @@ export default class IscnUploadForm extends Vue {
   @walletModule.Getter('getSigner') signer!: OfflineSigner | null
   @walletModule.Action('initIfNecessary') initIfNecessary!: () => Promise<any>
   @walletModule.Getter('getWalletAddress') address!: string
+  @bookApiModule.Getter('getToken') getToken!: string
```
nwingt, line 290: why name `getToken` instead `token` if this is not a function?

williamchong: it seems all getter is named as `getxxx` in this repo...

nwingt (approved):

PR #496, opened 2024-10-31
[TAS-2059] ✨ Add publisher and datePublished support

williamchong:
image


components/IscnRegisterForm.vue
```diff
@@ -1489,6 +1511,7 @@ estimation,
     }
     try {
       this.uploadStatus = 'signing'
+      console.log(formatISCNTxPayload(this.payload));
```
nwingt, line 1514: Please remove

nwingt (changes requested):

components/IscnRegisterForm.vue
williamchong:
```suggestion
```

nwingt (approved):

PR #512, opened 2025-02-05
[TAS-3201] 🚸 Display wallet info and enhance login behavior

AuroraHuang22:
截圖 2025-02-05 下午6 26 23

https://github.com/user-attachments/assets/87685ac9-682a-4217-8c54-31bb5bdab5fa


williamchong (approved):
Consider storing balance in store? Maybe you would need to restore it after sending transactions for Arweave/ISCN

